PR #7, opened 2018-12-20
Update issue templates

bmann:


bmann:
@pedrouid this is me experimenting with issue templates to help people submit in the right format. Probably just leave this for now, because we need to iterate on fields and format anyway.

ligi:
@bmann @pedrouid should we merge this now?

bmann:
@ligi I never used it before so wanted to experiment. Seems like a safe merge we can experiment with.

Issue #15, opened 2019-01-23
Use GH Pages instead of Netlify


ligi:
- [x] copy the just-the-docs theme in (GH Pages doesn't support it in remote theme mode)
- [x] change DNS to point to GitHub
- [x] change settings to build from GH Pages

@bmann refactor of root comment

@ligi reported SSL issues, which is a Netlify temporary issue with their CDN

---

I was just getting https errors from chainid.network

![image](https://user-images.githubusercontent.com/111600/51631445-0cd8fa00-1f4d-11e9-8391-a2e4cbed46b3.png)

After talking to @bmann it seems netifly is to blame as they also currently have API connectivity issues:

![selection_004](https://user-images.githubusercontent.com/111600/51631518-3e51c580-1f4d-11e9-8b23-75b530094e46.png)

I do not really see why we need netifly in the first place and would signal to just remove it.
cc @pedrouid

pedrouid:
I've just removed the API permissions

ligi:
wait - stop - it is still used currently
all for removing it - but we need to do it correctly - not just remove the api permissions

pedrouid:
But it's using Github Pages. Correct?
<img width="939" alt="screenshot 2019-01-23 22 14 30" src="https://user-images.githubusercontent.com/10136079/51640767-511fc680-1f5c-11e9-9106-e8f8255cff44.png">

ligi:
then we would not have had the fuckup with the ssl and netifly - afaik https://chainid.network is not the github pages but from netify - so we should not *yet* release it. First @bmann has to point https://chainid.network to github pages - then we can remove

bmann:
It's fine, I've renamed this issue and assigned to myself. No new builds will publish from commits until I've done this.

FYI -- as I said to Pedro, I used Netlify because the theme I was using isn't supported by GH Pages, and launched the site in like 15minutes :)

bmann:
OK, this should be done now. I have changed name servers to Namecheap, and [setup A records for chainid.network](https://help.github.com/articles/setting-up-an-apex-domain/) and turned on building from master branch to GH Pages. 

https will need to be enabled once GH recognizes that DNS has resolved.

Also, @pedrouid, I invited your email address to Namecheap so that you can manage DNS as well.

PR #16, opened 2019-01-25
Check in just-the-docs theme

bmann:
Addresses #15 

Have a look. GH Pages should build fine from master now, so as soon as this is merged, please go ahead and change settings.

Ping me and I can then change DNS.


ligi:
> Ping me and I can then change DNS.

@bmann ping

PR #20, opened 2019-02-08
Add Public RPC Endpoints

### LINKED ISSUE #12: Add RPC endpoints
wonder if we should add a list of RPC endpoints (can also be an empty list - so the field is optional) for the chains.
Inspired by https://github.com/ethereum-lists/chains/pull/10#issuecomment-454411114

This would also allow a ci-script to make checks for the chain_id (also inspired by the very same PR)

Disadvantage might be that it is widening the scope too much - what does the rest of the team think?

pedrouid:
Fix #12

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,15 +5,15 @@ Listed by chainId according to EIP-155
 Data source available on `_data/chains.json`
 
 Example Format
-```
+
+```json
 {
   "name": "Ethereum Mainnet",
   "short_name": "eth",
   "chain": "ETH",
   "network": "mainnet",
   "chain_id": 1,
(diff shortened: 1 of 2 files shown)


pedrouid:
Done, should we change the name then?

ligi:
yea - perhaps just "rpc": {...}

pedrouid:
Done! Good to merge

ligi:
:+1:

bmann:
Ideally all data fields are displayed on at least the /chains page which has enough room to include all information about the chain and can be human browseable.

minimayki6232 (approved):

Issue #25, opened 2019-02-10
ChainID 101 conflict


pedrouid:
There are two chains with the `chain_id` 101. We need to remove one of them!

```json
  {
    "name": "Webchain",
    "short_name": "web",
    "chain": "WEB",
    "network": "mainnet",
    "chain_id": 101,
    "network_id": 37129,
    "rpc": []
  },
  {
    "name": "EtherInc",
    "short_name": "web",
    "chain": "WEB",
    "network": "mainnet",
    "chain_id": 101,
    "network_id": 1,
    "rpc": []
  }
```

ligi:
I do not really want to be the judge of this. Can't there be both?

bmann:
It’s a really interesting case. I think informing both “owners” and helping them pick an un-used chain IDs is something we can help with. 

They don’t have to listen to us of course, but I think it’s the exact sort of thing that’s helpful for us to do. 

eg “before picking a ChainID, review ChainID.network and select an unused ChainID”

pedrouid:
It's no big deal as these two chain look the same, so it could be either a typo or a duplicate entry. 

There can only be one chain per chainId as per EIP-155 spec so we need to be the judges of conflicting scenarios like this. This is not the official list of EVM as there will never an official list but it's a trusted signal for listing EVM chains.

I checked and these two chains were part of the same commit:
https://github.com/ethereum-lists/chains/commit/9acc67d8b2f06c043be4b67e3a26690a8e026125

Hi @hackmod, could you review this chainId conflict and tell us where we could verify if this is a typo, duplicate or something else?

bmann:
Great. This is great info for the About page re: EIP-155 and to explain the goal here.

hackmod:
unfortunately, it is not a typo.
Please see https://github.com/kvhnuke/etherwallet/issues/2145

cc: @suryanshkumr @lukaszmatczak @mk

lukaszmatczak:
I agree that we need to resolve it somehow. But I think that root of this problem is inside EIP-155 specification. I've created issue about it: https://github.com/ethereum/EIPs/issues/1747

We need to wait and see how it will be fixed in specification and then resolve this conflict properly.

BTW. And of course there is a typo in list. EtherInc has "web" as "Short Name" and "Chain".

ligi:
> BTW. And of course there is a typo in list. EtherInc has "web" as "Short Name" and "Chain".

what should these be instead?

hackmod:
[...]

PR #28, opened 2019-02-12
Add new chain_id for Webchain

lukaszmatczak:
This change is to resolve issue #25 
I'll add correct block number when hard fork will be ready.

### DIFF
--- a/_data/chains.json
+++ b/_data/chains.json
@@ -265,6 +265,15 @@
     "rpc": []
   },
   {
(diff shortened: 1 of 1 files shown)


bmann:
What do you mean by changing the name?

ligi:
I just guess this:

` "name": "Webchain (after block xxxxxxx)",`

will not be permanent

pedrouid:
Agreed, let's merge for it now and then fix the name after the fork

PR #31, opened 2019-02-28
Added Lisinski testnet

Swader:


Swader:
It's pipa.lisinski.online, pipa is a Croatian bastardization of the word faucet :) All links (explorer etc) are available on the EN version of the landing page btw: https://lisinski.online/en

Roger that on the map, will do!

Issue #26, opened 2019-02-10
Create a Definitions page


bmann:
Create a field definitions page that explains each field, what it contains, and any usage / consumption details. 

This will help in people programmatically consuming the JSON file, since we can’t put comments in JSON.

pedrouid:
Related to discussion on #32

pedrouid:
Proposal following the discussion on #32: 

- name = name of the chain
- short_name = shortened and lowercase name
- consensus_type = blockchain consensus type
- native_currrency = blockchain native currency symbol
- ecosystem = blockchain ecosystem symbol
- network = name of the network related to cluster
- chain_id = number of the chain id
- network_id = number of the network id
- rpc = array of publicly available remote node rpc endpoints

I've copied the first draft that I commented previously, changed `chain` field to `ecosystem` and also added `consensus_type` and `native_currency`

PR #32, opened 2019-03-04
Fix xDAI submission

pedrouid:
Technically I've never really formalized what each parameter means and I kinda came up with a system that worked for my api aggregator: https://github.com/pedrouid/ethereum-api

However I hit an issue with how POA network describes xDAI chain as a POA chain named as DAI network, so I've included this in this PR. Let me know what you guys think

### DIFF
--- a/_data/chains.json
+++ b/_data/chains.json
@@ -186,8 +186,8 @@
   {
     "name": "xDAI Chain",
     "short_name": "xdai",
-    "chain": "XDAI",
-    "network": "mainnet",
+    "chain": "POA",
+    "network": "dai",
     "chain_id": 100,
     "network_id": 1,
     "rpc": ["https://dai.poa.network"]


bmann:
Yes this looks fine. Probably useful for us to define each element — essentially, write the “spec” for each label

ligi:
How would you define both fields now?

ligi:
@pedrouid bump

pedrouid:
The current state of EVM blockchains usually constitute multiple networks. The clients have the same implementation but different networks for different use-cases, most commonly a mainnet and a few testnets.

The relationship between these networks can be very subjective but the developer communities usually identify them as part of an aggregate or cluster. Therefore I think that this is aggregate would define what we call the chain field.

For Ethereum (ETH) we have multiple networks including mainnet, ropsten, rinkeby, kovan and goerli. These are all part of the ETH cluster, therefore they share the chain field `ETH` denominated by the ticker of the mainnet.

Same thing can be said about the POA network which are all part of a cluster of POA networks used as sidechains for Ethereum. These include core, sokol and xdai, therefore they share the chain field `POA`.

Hence the same reasoning can be followed for other chains. I know this is not a very technical unambiguous definition but it feels both accurate and practical for the current state of EVM blockchains.

The other fields are in my opinion self explanatory but here are my definitions:

- name = name of the chain
- short_name = shortened and lowercase name 
- chain = blockchain cluster ticker
- network = name of the network related to cluster
- chain_id = number of the chain id
- network_id = number of the network id
- rpc = array of publicly available remote node rpc endpoints

Additionally, I would like to propose an extra field for the chain's native currency denomination, some examples that could be useful in the current list: 

Ethereum Mainnet= ETH
Ethereum Ropsten = ropETH
Ethereum Rinkeby = rinETH
Ethereum Classic Mainnet = ETC
Ethereum Classic Testnet = tETC
POA Core = ETH
xDAI Chain = xDAI

pedrouid:
PS - I just realised I "renamed" in the last comment `xDAI Chain` to `POA xDAI`.
We should definitely drop the `Chain` word for starters but should we add `POA` to its name or would it be inaccurate? cc @ligi @bmann

ligi:
yea - we should drop chain as it is redundant - but I would not add POA

bmann:
POA is type of consensus so shouldn’t need to be in the name.

ligi:
@bmann yes - also how I see it - and we should really have a consensus field - and perhaps we should drop these for it:
```
chain = blockchain cluster ticker
network = name of the network related to cluster
```

really unsure what a cluster is in this context

pedrouid:
> POA is type of consensus so shouldn’t need to be in the name.

True but it's also a company/product/community of a network of POA sidechains

pedrouid:
> really unsure what a cluster is in this context

You could define a cluster by how the client dev teams coordinate with each other. ETH core devs coordinate releases with all clients for their own cluster (mainnet, ropsten, rinkeby, kovan, goerli) the POA network devs coordinate releases with all clients for their own cluster (core, sokol, xdai), the ETC core devs coordinate releases with all clients for their own cluster (mainnet, testnet).

Theoretically these are all EVM blockchains but they are separate dev communities with different roadmaps.

I find Blockscout's API endpoints make this clearer and IMO a good pattern to follow:

- Ethereum Mainnet => https://blockscout.com/eth/mainnet
- Ropsten Testnet => https://blockscout.com/eth/ropsten
- Rinkeby Testnet => https://blockscout.com/eth/rinkeby
- Kovan Testnet => https://blockscout.com/eth/kovan
- Goerli Testnet => https://blockscout.com/eth/goerli

- POA Core => https://blockscout.com/poa/core
- POA Sokol => https://blockscout.com/poa/sokol
- xDAI Chain => https://blockscout.com/poa/dai

- Ethereum Classic => https://blockscout.com/etc/mainnet

bmann:
Ah! I would call that ecosystem. And this is useful.

pedrouid:
I second the definition as `ecosystem`

pedrouid:
Should we merge this PR and move this dicussion to a new PR related to issue #26 "Create a Definitions page"

ligi:
I would signal not to merge it

"chain": "POA",
"network": "dai"


does not really make sense to me.

IMHO we should drop chain and network completely and replace by something "consensusType" and perhaps the proposed "ecosystem" (there I am not yet 100% convinced how it would be used but as @bmann thinks it is useful perhaps we add it)

pedrouid:
I agree that we should rename the `chain` field to `ecosystem` before merging but I wouldn't remove it.

Plus I also think that adding the `consensusType` would be useful 👍

pedrouid:
We will look into this later when we finish the Definitions page on #26

PR #34, opened 2019-03-11
Added ChainIDs for Metadium

MetadiumRelease:


ligi:
does the testnet have a faucet?

PR #37, opened 2019-03-20
Added PepChain Churchill testnet

MickdeGraaf:


ligi:
Do you have a faucet in this testnet?

PR #57, opened 2019-04-27
Updated network_id of High Performance Blockchain

Nicemanss:


ligi:
ah - you are right - was jumping lines there.

PR #70, opened 2019-06-17
Add verify script

pedrouid:
This script checks the following:
- file name matches the chainId
- chainId matches `eth_chainId` response
- networkId matches `net_version` response

Then it formats the JSON and overwrites the file with updated changes


pedrouid:
We could add verification for the schema and even ordering of the fields

PR #73, opened 2019-08-09
Added Xerom

fallengravity:
Name: Xerom
Ticker: XERO
Explorer: https://explorer.xerom.org
Public RPC: https://rpc.xerom.org
Website: https://xerom.org

Xerom is an Ethereum Fork with a modified Monetary policy, it also includes a new Masternode system which is run on-chain.


ligi:
:+1:

Issue #75, opened 2019-10-17
add slip44-id


ligi:
Needed e.g. for Multi-Coin ENS

PR #76, opened 2019-10-18
Support slip44 and initial backfill

ligi:
this is work towards #75 - don't think this is a complete backfill of chains that have a slip44 id yet - but I found no way to automate this - needed to manually cross-reference ..


ligi:
merging now - we can add more later

PR #78, opened 2019-11-06
Added clarification that "Ethereum Classic Testnet" is Ropsten.

bobsummerwill:


_data/chains/62.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "Ethereum Classic Testnet",
+  "name": "Ethereum Classic Testnet Ropsten",
```
soc1c: What? That is Morden...

bobsummerwill: Ha ha - brain fart.   I will make a new PR.

PR #79, opened 2019-11-06
Clarify that "Ethereum Classic Testnet" is Morden.

bobsummerwill:


ligi:
Does ETC has 2 testnets? Otherwise I rather not merge as it might lead to confusion with the original ETH testnet Morden

bobsummerwill:
It has multiple testnets, yes.

- Classic Morden
- Kotti (equivalent to Gorli)
- Mordor (new POW testnet, to replace Classic Morden, which has weird nonce thing only supported in Classic Geth, not Parity, not MultiGeth, not Besu)

@soc1c is the expert :-)

bobsummerwill:
The confusion with Morden is because it *is* the same testnet, which split at the time of the DAO fork.

bobsummerwill:
See https://ecips.ethereumclassic.org/ECIPs/ecip-1056 for the Meta ECIP for the pending Agharta HF, which includes block numbers for Morden, Mordor, Kotti and ETC mainnet.

ligi:
OK then - if there are multiple it makes sense

soc1c:
> Does ETC has 2 testnets? Otherwise I rather not merge as it might lead to confusion with the original ETH testnet Morden

It's the same network. ETH Morden = ETC Morden after ETH moved to Ropsten.

PR #80, opened 2019-11-07
add mordor classic testnet

soc1c:
https://github.com/eth-classic/mordor


ligi:
too bad there is no public RPC

soc1c:
not yet

PR #85, opened 2020-01-29
8(ubiq): update rpc, correct networkId

iquidus:


_data/chains/8.json
```diff
@@ -3,7 +3,7 @@
   "chain": "UBQ",
   "network": "mainnet",
   "rpc": [
-    "https://pyrus2.ubiqscan.io"
+    "https://rpc.octano.dev"
```
ligi: the old RPC seems to also still work - maybe we should leave it and add the other as a second? Or will the other one be disabled in the near future?

iquidus: There are no plans currently to phase out the old however the new endpoint is behind better infrastructure.  Can keep both but perhaps have the new one first?

ligi: Sounds good to me!

PR #94, opened 2020-05-08
Block 0xEB for Ethereum Beacon Chain

ligi:
The ChainId for the Beacon Chain is not decided yet - this PR is intended just to signal to other chains that this *migh* be the one

### DIFF
--- a/_data/chains/235.json
+++ b/_data/chains/235.json
@@ -1,0 +1,16 @@
+{
+  "name": "Ethereum Beacon Chain",
+  "chain": "EBC",
+  "network": "beaconchain",
+  "rpc": [],
+  "faucets": [],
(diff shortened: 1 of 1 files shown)


ligi:
another popular option is 0xE2 for Ethereum2.0

MysticRyuujin:
FWIW I think 0xEB makes more sense than 0xE2 given that the chain ID is for the Beacon Chain itself, assuming that shards will also require chainID.

Also an option: 0xBC 😄

roninkaizen:
👍

nehtechnine (changes requested):
Finish

ligi:
closing - not sure anymore beacon chain actually gets a chainID

PR #95, opened 2020-05-11
Create 122.json

LiorRabin:
Fuse mainnet


LiorRabin:
That was quick :) Thanks

PR #108, opened 2020-09-21
Update/add RPC addresses including WebSocket URLs.

saurik:
(Those two BSC endpoints are shown as http:// on their website, but are in fact https:// only.)

### DIFF
--- a/_data/chains/1.json
+++ b/_data/chains/1.json
@@ -4,7 +4,9 @@
   "network": "mainnet",
   "rpc": [
     "https://mainnet.infura.io/v3/${INFURA_API_KEY}",
-    "https://api.mycryptoapi.com/eth"
+    "wss://mainnet.infura.io/ws/v3/${INFURA_API_KEY}",
+    "https://api.mycryptoapi.com/eth",
+    "https://cloudflare-eth.com"
   ],
   "faucets": [],
   "nativeCurrency": {
--- a/_data/chains/100.json
+++ b/_data/chains/100.json
@@ -3,7 +3,13 @@
   "chain": "XDAI",
   "network": "mainnet",
   "rpc": [
-    "https://dai.poa.network"
+    "https://rpc.xdaichain.com",
+    "https://xdai.poanetwork.dev",
+    "wss://rpc.xdaichain.com/wss",
+    "wss://xdai.poanetwork.dev/wss",
+    "http://xdai.poanetwork.dev",
+    "https://dai.poa.network",
+    "ws://xdai.poanetwork.dev:8546"
   ],
   "faucets": [],
   "nativeCurrency": {
--- a/_data/chains/3.json
+++ b/_data/chains/3.json
@@ -3,7 +3,8 @@
   "chain": "ETH",
   "network": "ropsten",
   "rpc": [
-    "https://ropsten.infura.io/v3/${INFURA_API_KEY}"
+    "https://ropsten.infura.io/v3/${INFURA_API_KEY}",
+    "wss://ropsten.infura.io/ws/v3/${INFURA_API_KEY}"
   ],
   "faucets": [
     "https://faucet.ropsten.be?${ADDRESS}"
--- a/_data/chains/4.json
+++ b/_data/chains/4.json
@@ -3,7 +3,8 @@
   "chain": "ETH",
   "network": "rinkeby",
   "rpc": [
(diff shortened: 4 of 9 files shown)


_data/chains/43110.json
```diff
@@ -6,7 +6,7 @@
     "https://ava.network:21015/ext/evm/rpc"
   ],
   "faucets": [
-    "http://athfaucet.ava.network//?address=${ADDRESS}"
+    "https://faucet.avax-test.network"
```
ligi, line 9: wonder if this is actually the same chain - seems they discontinued the old one?!

ligi (commented):
Thanks so much for the PR - looks good - just not sure about the ava one.

_data/chains/43110.json
saurik: So, the faucet URL I provided lets you get tokens on Avalanche's Ethereum-compatible C-chain--which has chain ID 43110--if you prefix your address with "C-". The faucet URL that was there before is for a website that doesn't even exist? ;P However, you are also correct, in that I somehow failed to really read the rest of this file, which is claiming this is for Avalanche Athereum, which doesn't really exist right now? I'm going to poke them about this chain ID situation... :/. In the meantime, I've simply removed this change (leaving this file as it was, with the definitely-wrong data from before instead of my replacement maybe-inconsistent data) and pushed a new commit.

ligi (approved):

_data/chains/43110.json
saurik: @ligi So, I talked with the Avalanche people, and they now have this file of constants they sent me (which was apparently just updated with correct chainids yesterday ;P). In all cases this is for the C-chain of said network. (There are networks listed for 2 and 3 but no associated chainid.)

https://github.com/ava-labs/avalanchejs/blob/development/src/utils/constants.ts

ChainID NetworkID Names
43111 0 Manhattan
43114 1 Avalanche, Mainnet
43110 4 Everest
43113 5 Fuji, Testnet

saurik: Oh, and 43112 is used for "custom" (local) networks as a fallback.

ligi: yea - but it seems they broke ethereum compatibility - e.g. other address format. So IMHO we should not add them to chainID network

saurik: @ligi They did? FWIW, I have been testing stuff against them and the only thing I have run into that isn't Ethereum compatible is that they have a hardcoded minimum gas of 470 Gwei (which prevents deployment of some standardized singleton contracts, such as EIP-1820). They are using Geth to do all the EVM stuff. If you are referring to my mention of C- prefix on that faucet, that is because they have another chain called the X-chain that you can swap tokens to, and so the faucet wants you to tell it if you want to get the money to the X or C chains, but the Avalanche C-chain is designed to be "byte, byte, byte compatible with Ethereum", and they even quickly arranged to upgrade to Istanbul (or later) for me as part of their mainnet launch.

https://twitter.com/avalabsofficial/status/1305249913723330560?s=21

saurik: @ligi Their JSON/RPC endpoint seemed fully compatible with my expectations as well (which is because I believe/assume it is being handled by Geth).

ligi: yea - was also really surprised (and annoyed)  - the last testnet worked
fine. Really hope I am wrong - just see here:
https://twitter.com/mr_ligi/status/1308020223296106496

saurik: @ligi Well, that faucet UI doesn't have a GUI selector for chain, so you do need to type "C-" before the 0x. That is just how the faucet works, not how their chain functions ;P.

Also, right now, their network is not even online, as they took Everest down yesterday to bring up their mainnet today; I have no clue what the behavior for that faucet (which is often broken anyway) thereby might be with the backing network missing all of its bootstrap nodes.

Like, _right now_, their underlying chain explorer is just a coming soon dialog, and their C-chain explorer page is confused with no blocks ;P. In another few hours, I would expect it all to be back up again, with the various web pages fixed and operational.

https://explorer.avax.network/
https://cchain.explorer.avax.network/

ligi: cool - that is good news actually. Was just a bit confused by - https://github.com/ava-labs/avax-js-cli-tools/blob/master/eth_convert.js
Just staying patient now until the dust settles ..-)

PR #109, opened 2020-09-21
Add Optimistic Eth

karlfloersch:
Here's the submission! One question: I set `"chain": "ETH",` -- is that what you were looking for? This is a proper L2 so from a security model it is the same chain as ETH but was just curious what the field means. 

Anyway, hope this works!

### DIFF
--- a/_data/chains/420.json
+++ b/_data/chains/420.json
@@ -1,0 +1,16 @@
+{
+  "name": "Optimistic Ethereum",
+  "chain": "ETH",
+  "network": "mainnet",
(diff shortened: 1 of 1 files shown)


ligi (approved):

ligi:
Nice! It's the first L2 chain - maybe we should add some more L2 specific stuff later (I could e.g. imagine the chainId that it is anchored to - so we get a chain of chains. And maybe you have other ideas on what we should add L2 specific)
Thanks for the PR - hope to try it out soon!

PR #110, opened 2020-10-05
Adding NEAR EVM

ilblackdragon:
NEAR EVM behaves like Ethereum shard on NEAR and support set of tools like Truffle / Metamask, hence the chain-id requirement.

Details about NEAR EVM can be found in WIP spec: https://github.com/nearprotocol/NEPs/pull/106

The id for MainNet is hex("NEAR")

### DIFF
--- a/_data/chains/1313161554.json
+++ b/_data/chains/1313161554.json
@@ -1,0 +1,19 @@
+{
+  "name": "NEAR MainNet",
+  "chain": "NEAR",
+  "network": "mainnet",
+  "rpc": [
+    "https://rpc.mainnet.near.org"
+  ],
+  "faucets": [
+  ],
+  "nativeCurrency": {
+    "name": "NEAR",
+    "symbol": "NEAR",
+    "decimals": 24
+  },
+  "infoURL": "https://near.org/",
+  "shortName": "near",
+  "chainId": 1313161554,
+  "networkId": 1313161554
+}
--- a/_data/chains/1313161555.json
+++ b/_data/chains/1313161555.json
@@ -6,6 +6,7 @@
     "https://rpc.testnet.near.org"
   ],
   "faucets": [
+    "https://wallet.testnet.near.org"
   ],
   "nativeCurrency": {
     "name": "NEAR",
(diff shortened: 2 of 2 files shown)


ligi:
Thanks for the PR - are there really no faucets for the testnet?

ilblackdragon:
Added faucet for Testnet.

mikedotexe:
@ligi a team member at NEAR pointed out that the "Short Name" is the same for both NEAR testnet and NEAR mainnet, will this cause issues?

saurik:
Is this really an Ethereum-compatible chain? I was excited to see it added, but the RPC endpoint doesn't seem to respond to any Ethereum methods. I've seen endpoints that don't support eth_chainId before (Cloudflare, for example) and I was annoyed but not too put off by eth_gasPrice not working... but, when eth_blockNumber didn't even work, I was left scratching my head as to what the listed RPC endpoint is accomplishing :(.

ilblackdragon:
@saurik No, RPCs are not Ethereum compatible.

There is a proxy web3 Provider here - https://github.com/near/near-web3-provider There will be also RPC proxy, later - there is an open bounty for building it.

Also the work is still happening to nail down the protocol update of adding EVM here - https://github.com/near/nearcore/pull/3257, before it's going to make to TestNet / MainNet.

This PR was done ahead of time, so we can start testing tooling, like MetaMask and other tools to recognize NEAR.

saurik:
@ilblackdragon If the RPC endpoint isn't Ethereum compatible then I'd argue it really shouldn't be listed in the rpc field of this Ethereum chain database, as it isn't actually usable or compatible with any of the other usages of the field. The expectation should be that I can, given just a chain's ID, pull its RPC endpoint from this database and talk to it (and to that end one of the things I've been working on a bit is automated testing of all the RPC endpoints in this database by doing a quick minimal check that, at least, calling eth_chainId returns the expected chain, which is important as sometimes these networks--I'm looking at you, Avalanche ;P--messes with their chains a bit while lauching).

ilblackdragon:
@saurik Good point, let me send a PR to remove RPC endpoints for now.

pedrouid:
btw @ilblackdragon you should also consider introducing a CAIP for Near... here is an example of a current PR for EOSIO namespace (ChainAgnostic/CAIPs#33)

Issue #4, opened 2018-12-20
import from Trezor ethereum networks list


prusnak:
For Trezor we maintain a similar list here: https://github.com/trezor/trezor-common/blob/master/defs/ethereum/networks.json

You might want to add new networks from the list and also add the fields from our list. Except for `blockbook` all of them are not-specific for Trezor and could be useful for everybody.

Feel free to ask for details if needed.

ligi:
thanks - very helpful!

pedrouid:
Thanks a lot! I added some chains from that list, I will revisit this about the other fields.

ligi:
closed via #117

PR #117, opened 2020-11-06
fill missing SLIP44 IDs and some misc updates

matejcik:
updating the chains repo from Trezor's Ethereum network data, so that we can switch to using your repo as a source


matejcik:
fixes #4 apparently

ligi:
oh thanks!

PR #128, opened 2020-11-20
adding Meter mainnet chain id

youngmou:
Adding the Meter mainnet chainID, which is set as 82 (0x52)

### DIFF
--- a/_data/chains/82.json
+++ b/_data/chains/82.json
@@ -1,0 +1,19 @@
+{
+  "name": "Meter Mainnet",
+  "chain": "METER",
+  "network": "mainnet",
+  "rpc": [
+    "https://rpc.meter.io",
+  ],
+  "faucets": [],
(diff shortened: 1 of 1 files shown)


_data/chains/82.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "Meter Mainnet",
+  "chain": "METER",
+  "network": "mainnet",
+  "rpc": [
+    "https://rpc.meter.io",
```
ligi, line 6:
```suggestion
    "https://rpc.meter.io"
```

ligi (changes requested):
Please remove this extra comma

_data/chains/82.json
youngmou: done as suggested.

ligi (approved):

PR #132, opened 2020-11-25
Use CAIP-2 as filename

### LINKED ISSUE #130: Use CAIP-2 as filename
currently we use EIP-155 for the filename - but this can lead to collisions (or better a problem adding chains) as @matejcik pointed out here: https://twitter.com/yangWao/status/1330871940534448128

A solution can be to use CAIP-2 as basis for the filename

Please veto now if you see problems ahead with this!

cc @pedrouid @bmann

pedrouid:
Fix #130


pedrouid:
I had to use `eip155-x` instead of `eip155:x` for the file system

ligi:
thanks - will need to adapt the CI so it is happy
but one question first - where do you see problems with the ":" so the transform is needed?
I just tried it out and git and linux seems to be ok with ":" in filenames:

```
✓ ~/t/caip (master +%) git st                                           22:10:07
On branch master

No commits yet

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	eip155:10

nothing added to commit but untracked files present (use "git add" to track)

```

pedrouid:
I used the macOS Finder to quickly rename all files and it didn’t allow me

ligi:
can you let me push to this branch?

ligi:
or cherry pick this commit: https://github.com/ethereum-lists/chains/commit/fb7db673f8ea3899265db122e779f34cd8475f23

fixes the CI

ligi:
@matejcik can you approve or weigh in if this change works for TREZOR?

matejcik:
this is fine for us

ligi:
@pedrouid any reason it is still in draft from your side?
can you cherry pick this commit: fb7db67

pedrouid:
Sorry @ligi for the delay. I've changed the PR status to ready and cherry picked your commit 👍

ligi:
something must have went wrong with the cherry-pick. It does not reflect all changes from the commit. The renaming of the resources is missing - hence all tests fail

ligi:
@pedrouid *push* - I can also take it over if you want - but we need to get this going as things diverge now with new incoming chains

pedrouid:
@ligi  perhaps it's better that you handle it since you have more experience with Gradle

PR #133, opened 2020-11-27
add MathChain Network

hging:


ligi (changes requested):
the shortNames can't be the same - it's a constraint of TREZOR

hging:
> the shortNames can't be the same - it's a constraint of TREZOR

I change the testnet short name to tMATH.

ligi (approved):

PR #141, opened 2020-12-15
Add Support for XinFin[XDC] Network

akhekade:
Adding Support for XinFin
https://coinmarketcap.com/currencies/xinfin-network/markets/


ligi (changes requested):
shortName and name must be unique

ligi (approved):

PR #142, opened 2020-12-21
add Huobi ECO Chain info

wangsai-silence:
add info for Huobi ECO Chain


ligi (changes requested):
the shortName must not be the same

wangsai-silence:
> the shortName must not be the same

I renamed it. Please review. Thank you.

PR #143, opened 2021-01-04
add okexchain info

ilovers:


ligi (changes requested):
The shortNames must not be the same

ilovers:
> The shortNames must not be the same

Thank you for your prompt reply. I have renamed the shortName. Please review.

ligi (approved):

PR #146, opened 2021-01-15
Update Optimistic Ethereum ChainID for mainnet

karlfloersch:
We're shipping a fresh chain to mainnet tomorrow & therefore have to change the chainID. I believe you suggested this when I submitted my first PR but we foolishly did not heed your advice. It seems 420 for mainnet was not meant to be.

Updated 420 to forever be our Goerli testnet & now 10 as the mainnet!

Thanks again for maintaining this repo @ligi 😊


ligi (approved):
:+1:

PR #148, opened 2021-01-17
Remove faucets for optimistic goerli

ligi:
These faucets are for goerli - but not for optimistic goerli as far as I see. Currently this leads to weird UX in WallETH as after you request funds from the faucet you do have them on the current chain:

image

Wondering if there are faucets for optimistic goerli currently

### DIFF
--- a/_data/chains/eip155-420.json
+++ b/_data/chains/eip155-420.json
@@ -3,10 +3,7 @@
   "chain": "ETH",
   "network": "goerli",
   "rpc": [],
-  "faucets": [
-    "https://goerli-faucet.slock.it/?address=${ADDRESS}",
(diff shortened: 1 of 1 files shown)


ligi:
This slipped when reviewing https://github.com/ethereum-lists/chains/pull/146/files - @karlfloersch would like your input on this

K-Ho:
@ligi - I think we were thinking that users can use the L1 Goerli ETH faucet and then deposit them into L2, so we just listed the L1 Goerli ETH faucet links there. Definitely think you're right we should remove those faucet links, since it seems it could be misleading!

ligi:
@K-Ho thanks for the info!

PR #153, opened 2021-01-27
Add ChainID for SoterOne

scott-ss:
SoterOne uses secure multi-party computation (MPC), differential privacy, machine learning and
blockchain technology to create a platform that integrates privacy protection,
data modeling and data sharing.

SoterOne uses the following chain ID:
- 218 (0xda) for SoterOne Mainnet


Angie3120 (commented):
@

PR #154, opened 2021-02-02
Change SoterOne chain network id to 68

scott-ss:
This patch changes the SoterOne chain id to 68.

Testing with a Ledger hardware wallet plus Metamask configuration revealed that the previous chainId, 218, fails with "Ledger: The ransaction signature is not valid" even with the latest firmware (1.6.1).


ligi (changes requested):
Is the old one (with id 218) stopping then? This change just opens a second one - then they also need different ShortNames

scottt:
> Is the old one (with id 218) stopping then? This change just opens a second one - then they also need different ShortNames
@ligi,
1. The old chain with id 218 is stopping. 
2. This patch is a Git rename, which I think properly signals the old chain is stopping:

```
diff --git a/_data/chains/eip155-218.json b/_data/chains/eip155-68.json
similarity index 89%
rename from _data/chains/eip155-218.json
rename to _data/chains/eip155-68.json
```

scott-ss:
> Is the old one (with id 218) stopping then? This change just opens a second one - then they also need different ShortNames

@ligi , 
1. The chain with chain id 218 is stopping
2. This patch is a git rename, which I thought was properly signaling that:

```
diff --git a/_data/chains/eip155-218.json b/_data/chains/eip155-68.json
similarity index 89%
rename from _data/chains/eip155-218.json
rename to _data/chains/eip155-68.json
index 62ee4bf..8b9d9c3 100644
```

ligi:
Seems to be no rename - seems to be a addition:

Selection_458

Also can you move the old one into a folder "deprecated"? I think we should preserve this so using apps see that a cain stopped

scott-ss:
> Can you move the old one into a folder "deprecated"?

Done in https://github.com/ethereum-lists/chains/commit/af854c696d1a50042bac6d72eb523f6a3c5b3342

scott-ss:
@ligi, Re: file rename vs addition

The change is clearly sown as a file rename here:
https://github.com/ethereum-lists/chains/pull/154/commits/af854c696d1a50042bac6d72eb523f6a3c5b3342

PR #155, opened 2021-02-03
add harmony mainnet chain IDs

rlan35:
Harmony is evm-compatible and sharded blockchain where each shard can act as a individual ethereum-compatible chain. 

The chainIDs we reserved for the 4 shards in Harmony mainnets are:

shard 0: 1666600000
shard 1: 1666600001
shard 2: 1666600002
shard 3: 1666600003

### DIFF
--- a/_data/chains/eip155-1666600000.json
+++ b/_data/chains/eip155-1666600000.json
@@ -1,0 +1,19 @@
+{
+  "name": "Harmony Mainnet Shard 0",
+  "chain": "Harmony",
+  "network": "mainnet",
+  "rpc": [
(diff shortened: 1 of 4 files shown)


ligi (changes requested):
the shortNames must be different between chains
Also: do these shards have some chain they are anchored in in our list of chains?

rlan35:
> the shortNames must be different between chains
> Also: do these shards have some chain they are anchored in in our list of chains?

fixed. Our sharded chain have our own unique chainID which is shared among all shards. But for ethereum-compatibility, we  have to use separate chain ID to identify the different shards.

ligi (approved):

PR #159, opened 2021-02-12
Fix RPC link of TomoChain

boringcrypto:
Source: https://docs.tomochain.com/developer-guide/working-with-tomochain/tomochain-mainnet


ligi:
thanks!

PR #167, opened 2021-02-24
Introduce the devETH network to the chain list.

vkoskiv:


ligi (approved):

ligi (changes requested):
The filename must match the chainID
currently it is matching the networkID (are they really this way or is thee a typo?)

vkoskiv:
Yeah, I just spotted that. Looks like there's already a 787 on the list there for Acala. Any workaround?

vkoskiv:
No typo, the networkID is 777 and chainID is 787, indeed.

ligi:
Nope - there is no workaround. Duplicate chainIDs are bad - not allowed.

vkoskiv:
Fair enough. Thanks for taking a look anyway!

ligi:
change your chainID now that you are still a fresh project. Will become harder and harder in the future!

PR #168, opened 2021-02-25
Add darwinia test network

AsceticBear:


ligi (changes requested):
There can only be one native currency - wonder why you need 2

AsceticBear:
> There can only be one native currency - wonder why you need 2

Let me explain the details more. Darwinia network is a decentralized cross-chain bridge based on substrate framework. Due to the nature of substrate, we support two currencies now, RING, KTON.

The Pangolin network is a test network built by Darwinia to support ethereum evm which allows users to migrate existing Dapps to the Darwinia network without much modification.

ligi:
thanks for the info! But when a (EVM) smart-contract is executed - from which of the currencies does it take the gas?

AsceticBear:
> thanks for the info! But when a (EVM) smart-contract is executed - from which of the currencies does it take the gas?

The first one, RING.

ligi:
please set RING as native currency then. We can also add a new field maybe for a secondary currency. But of that I would need more information how this information can be used then.

AsceticBear:
> please set RING as native currency then. We can also add a new field maybe for a secondary currency. But of that I would need more information how this information can be used then.

Done, Please check again later.

_data/chains/eip155-43.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "Darwinia Pangolin Testnet",
+  "chain": "pangolin",
+  "network": "free testnet",
+  "rpc": [],
+  "faucets": [],
+  "nativeCurrency1": {
```
ligi, line 7:
```suggestion
  "nativeCurrency": {
```

ligi (changes requested):

ligi:
I am happy with it now - but CI is missing the shortName field

PR #169, opened 2021-03-03
Add clover network ids

wm-chen:
clover network


ligi (changes requested):
Please add the shortName field

ligi (changes requested):
the shortNames of both chains must be different

ligi (approved):

PR #187, opened 2021-03-16
Add chain-id 78 for our mainnet

fbaud:
Following the testnet we added with id 78110, we launch our mainnet with id 78


ligi (changes requested):
org.ethereum.lists.chains.model.NameMustBeUnique:  name must be unique - but found `Firenze test network` more than once

fbaud:
Removed the offending file in older naming format. Sorry for the error.

ligi (approved):

PR #189, opened 2021-03-18
update okexchain rpc

ilovers:


_data/chains/eip155-65.json
```diff
@@ -3,12 +3,12 @@
...
   "network": "testnet",
   "rpc": [
-    "https://www.okex.com/okexchain-rpc-testnet"
+    "https://exchaintest.okexcn.com"
   ],
   "faucets": ["https://www.okex.com/drawdex"],
   "nativeCurrency": {
     "name": "OKExChain Global Utility Token in testnet",
-    "symbol": "TOKT",
+    "symbol": "OKT",
```
ligi, line 11: I think it is good that the testnet has a different symbol than the mainnet. Don't you think?
Rest of the PR looks good

ilovers: > I think it is good that the testnet has a different symbol than the mainnet. Don't you think?
> Rest of the PR looks good

Yes, I agree with you, but we have lanuched testnet with OKT symbol.
Thank you for your kind help.

ligi (approved):

PR #190, opened 2021-03-18
add id 22 23 for elastos did sidechain mainnet and testnet

JeanMaritain:


JeanMaritain:
thank you, will assign unique short name.

PR #203, opened 2021-04-02
Update Optimistic Ethereum chain info

K-Ho:
Minor updates to Optimistic Ethereum chain info :)

### DIFF
--- a/_data/chains/eip155-10.json
+++ b/_data/chains/eip155-10.json
@@ -6,7 +6,7 @@
   "faucets": [],
   "nativeCurrency": {
     "name": "Ether",
-    "symbol": "OETH",
+    "symbol": "ETH",
     "decimals": 18
   },
   "infoURL": "https://optimism.io",
--- a/_data/chains/eip155-69.json
+++ b/_data/chains/eip155-69.json
@@ -1,12 +1,12 @@
 {
-  "name": "Optimistic Ethereum Testnet Kovan",
+  "name": "Optimistic Kovan",
   "chain": "ETH",
   "network": "kovan",
   "rpc": ["https://kovan.optimism.io/"],
   "faucets": [],
   "nativeCurrency": {
     "name": "Kovan Ether",
-    "symbol": "KOR",
+    "symbol": "ETH",
     "decimals": 18
   },
   "infoURL": "https://optimism.io",


ligi (changes requested):
I am OK with the name change but do not really like the renaming of the symbol. IMHO it is good that the symbols are distinct

l2ichshop2 (approved):

8214246823 (changes requested):
[K-Ho:master]()

K-Ho:
Hey @ligi, appreciate the review. RE:native currency symbols, ETH that has been deposited into Optimistic Ethereum is referred to everywhere as ETH (in wallets, dapps, explorers, etc). Having the symbol listed as OETH in the ethereum-lists right now means that either 
1. Users get an unsavory error message whenever adding the network in their metamask (see below) or 
2. We start setting the symbol for new users as OETH which would cause a ton of confusion between users that are shown as having OETH and the users that are shown as having ETH. 

If we can't change the symbol in this repo, then maybe we can instead work together with Metamask to not check that the symbol being added for the network is the same as what is listed in this repo. One thing I noticed is that Kovan ETH is listed as having a symbol of KOR in this repo, but Metamask displays Kovan ETH with a symbol of ETH, so there does seem to be a precedent with Metamask diverging from this list. Would love to hear your thoughts on this issue!

ligi:
sorry for the delay - missed the notification about this.
IMHO it really should not be ETH - and I am not the only one thinking so - made a poll about this once: https://twitter.com/mr_ligi/status/1393135790948036610
and it is just a different thing with different trust assumptions - so it should be called differently.
Really think this should be fixed on Metamasks side and not here.
Also in the future wallets will need to go away from a network switcher and display assets across networks - what do you do then - so you may have X ETH and Y ETH - really confusing - better you have X ETH and Y OETH

bdresser:
Hello @ligi!

As @K-Ho mentioned, applications and wallets have adopted a uniform treatment here in calling the asset ETH regardless of chain (shown below). MetaMask is trying to keep folks safe by making sure network details match; I think updating chain info here is better for users than asking MM to remove a safeguard. Consistency across L2 is key in making Ethereum accessible, clear, and affordable! 

Lmk what you think or if you'd like to chat live. 

<br>

Uniswap:
Screen Shot 2022-02-04 at 6 20 33 PM

Lyra: 
Screen Shot 2022-02-04 at 6 19 59 PM

Rainbow:


Coinbase Wallet:

ligi:
> MetaMask is trying to keep folks safe by making sure network details match; I think updating chain info here is better for users than asking MM to remove a safeguard.

no they should not remove the safeguard - they should change the Symbol ..-)
I really think it is a mistake - but seems all wallets do this mistake - so maybe we really need to merge that mistake. But for this the branch must be rebased.

bdresser:
Thanks @ligi! I opened a clean PR here: https://github.com/ethereum-lists/chains/pull/843

alirezaayande (approved):
_data/chains/eip155-10.json

alirezaayande (approved):
gh pr checkout 203

alirezaayande:
> gh pro تسویه حساب 203

ligi:
replaced by #843

PR #215, opened 2021-04-13
add quarkchain chainids

ping-ke:
add quarkchain chainids
    mainnet: https://mainnet.quarkchain.io/ 
    devnet: https://devnet.quarkchain.io/

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -69,4 +69,4 @@ DEPENDENCIES
   tzinfo-data
 
 BUNDLED WITH
-   1.16.3
+   2.1.4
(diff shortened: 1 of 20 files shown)


ligi (changes requested):
I think the RPCs are wrong - they all point to the same
Also these are a lot of shards - will need to check with consumers - we need a way to fold here

l2ichshop2 (approved):

ping-ke:
> I think the RPCs are wrong - they all point to the same
> Also these are a lot of shards - will need to check with consumers - we need a way to fold here

ping-ke:
All shards rpc calls are implemented by calling root chain rpc, so they all fill in the same PRC.  is that work?
For "will need to check with consumers - we need a way to fold here", do you mean we need to create a folder like qkcchainids, under "_data/chains" and move all those files to that folder?

ligi:
Not yet sure - how do you target the shards in the calls then? I can Imagine for transactions via EIP155 - but how does this work for other calls?

PR #219, opened 2021-04-14
Update eip155-269.json

wgllele:
Update to correct address


ligi (approved):

ligi:
you might also want to fill out the explores field as outlined in the README

PR #222, opened 2021-04-15
Update eip155-269.json

wgllele:
Update to the correct information.


ligi (commented):
Why did you remove slip-44?

ligi (changes requested):
Should not remove the SLIP44

Hubtig23:
This should be used as reference material it have predictable and constant environment deployment as a gain

Hubtig23:
#229

PR #229, opened 2021-04-25
b_e_b_chain plus Mstr

Hubtig23:
Splicing Cross-Chain features successfully is the goal will become a contribution


ligi (changes requested):
You need to target the master brach

PR #230, opened 2021-04-25
update okexchain infoURL

ilovers:


ligi (changes requested):
I think the infoURLs are correct  but the other ones should be added as explorers like you can see in the readme

ilovers:
> I think the infoURLs are correct but the other ones should be added as explorers like you can see in the readme

Yes, you are right. I will close this pr, and reopen a new one if we need. Thanks for your kind help :)

ilovers:
> I think the infoURLs are correct but the other ones should be added as explorers like you can see in the readme

We considered, we still need to update infoUrl to https://www.oklink.com/okexchain. Because other tools use infoUrl as explorer url. Thanks again for your help.

ligi:
> Because other tools use infoUrl as explorer url. 

which tools are you referring here?

ilovers:
> which tools are you referring here?

For example, [chainlist.org](https://chainlist.org/), when user use the "Add Your Network" feature to add the okexchain network to MetaMask, it used infoURL as the Block Explorer URL.

ligi:
There is an issue upstream https://github.com/antonnell/networklist-org/issues/12 - hope it get's fixed soon. But changing it here is the wrong approach.

PR #233, opened 2021-04-26
add KuCoin Community Chain(KCC) testnet

lingyun92:


ligi (changes requested):
The exporers section is invalid JSON
Also the exporer does not adhere EIP3091 (block is broken - https://scan-testnet.kucoin.one/block/32735)
So they either need to change this or create their own standard

lingyun92:
> The exporers section is invalid JSON
> Also the exporer does not adhere EIP3091 (block is broken - https://scan-testnet.kucoin.one/block/32735)
> So they either need to change this or create their own standard

fixed, thanks.

ligi:
Great! Seems the explorer is close to following the standard now - only tokens is still wrong as far as I see.
https://scan-testnet.kucoin.one/token/0xb296bab2ed122a85977423b602ddf3527582a3da

Also please add the  "standard": "EIP3091" then - after I can merge

lingyun92:
> Great! Seems the explorer is close to following the standard now - only tokens is still wrong as far as I see.
> https://scan-testnet.kucoin.one/token/0xb296bab2ed122a85977423b602ddf3527582a3da
> 
> Also please add the "standard": "EIP3091" then - after I can merge

works now

_data/chains/eip155-322.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "KuCoin Community Chain Testnet",
+  "chain": "KCC",
+  "network": "testnet",
+  "rpc": [
+    "https://rpc-testnet.kucoin.one",
```
ligi, line 6:
```suggestion
    "https://rpc-testnet.kucoin.one"
```

ligi (changes requested):
CI found this extra comma and complained about it

ligi (approved):

ligi:
Long time in the making - but finally merged ;-)

PR #239, opened 2021-04-30
Fix on Fantom Public RPCs.

firke:
In previous PR did not look into the RPC values, at the moment only working public one is https://rpcapi.fantom.network, so that's the only one in the list now.
References to it:
https://ftmscan.com/apis#rpc
https://docs.fantom.foundation/tutorials/set-up-metamask


ligi:
:+1:

PR #247, opened 2021-05-04
Add Binance Smart Chain explorer

zepxtor:


_data/chains/eip155-56.json
```diff
@@ -26,5 +26,10 @@
...
   "shortName": "bnb",
   "chainId": 56,
-  "networkId": 56
-}
\ No newline at end of file
+  "networkId": 56,
+  "explorers": [{
+    "name": "bscscan",
+    "url": "https://bscscan.com",
+    "icon": "bscscan"
```
ligi, line 33:
```suggestion
    "icon": "bscscan",
    "standard": "EIP3091"
```

ligi (changes requested):
Please add the standard field as suggested. The explorer seems to actually respect the EIP (which is good). 
If you add the icon like this you also need to add a icon-definition as described in the readme

ligi (approved):

PR #248, opened 2021-05-04
Add Binance Smart Chain Testnet explorer

zepxtor:


_data/chains/eip155-97.json
```diff
@@ -21,5 +21,10 @@
   "infoURL": "https://testnet.binance.org/",
   "shortName": "bnbt",
   "chainId": 97,
-  "networkId": 97
+  "networkId": 97,
+  "explorers": [{
+    "name": "bscscan-testnet",
+    "url": "https://testnet.bscscan.com",
+    "icon": "bscscan"
```
ligi, line 28:
```suggestion
    "icon": "bscscan",
    "standard": "EIP3091"
```

ligi (changes requested):
Please add the standard field as suggested. The explorer seems to actually respect the EIP (which is good). 
If you add the icon like this you also need to add a icon-definition as described in the readme

ligi (approved):

PR #249, opened 2021-05-05
add explorer for matic

Code0x2:
Adds [https://explorer-mainnet.maticvigil.com](https://explorer-mainnet.maticvigil.com) for matic as explorer


ligi (changes requested):
Unfortunately it is not really 3091 compatible - it fails for block and token (wants plural in both cases)

PR #254, opened 2021-05-10
Add Lucid Network

mikemaid:
https://lucidcoin.io


ligi (changes requested):
This json is invalid - one JSON can only contain one chain

mikemaid:
2 Networks can't exist on the same chain?

ligi:
technically they can - but we only accept the oldest one as otherwise replay attacks are possible

PR #256, opened 2021-05-10
Boba Network

InoMurko:


_data/chains/eip155-28.json
```diff
@@ -0,0 +1,22 @@
...
+    "symbol": "ETH",
+    "decimals": 18
+  },
+  "icon": "omgx",
+  "infoURL": "https://www.omgx.network/",
+  "shortName": "omgx",
+  "chainId": 28,
+  "networkId": 28,
+  "explorers": [{
+    "name": "etherscan",
```
ligi, line 18: pretty sure this is not etherscan ;-)
also the block explorer did not load for me - so I could not check if it is really EIP3091 compatible

ligi (changes requested):

_data/chains/eip155-28.json
InoMurko: whooops, indeed!
@ligi block explorer should be up now, thank you for checking!

PR #259, opened 2021-05-13
Update the NEAR chains

artob:
[EVM on NEAR](https://github.com/aurora-is-near/aurora-engine) is now called [Aurora](https://github.com/aurora-is-near), and has launched as of yesterday:

- https://near.org/blog/aurora-launches-near/
- https://aurora.dev
- https://doc.aurora.dev/develop/networks

### DIFF
--- a/_data/chains/eip155-1313161554.json
+++ b/_data/chains/eip155-1313161554.json
@@ -8,7 +8,7 @@
   "faucets": [],
   "nativeCurrency": {
     "name": "Ether",
-    "symbol": "ETH",
+    "symbol": "aETH",
     "decimals": 18
   },
   "infoURL": "https://aurora.dev",
--- a/_data/chains/eip155-1313161554.json
+++ b/_data/chains/eip155-1313161554.json
@@ -1,18 +1,18 @@
 {
-  "name": "NEAR MainNet",
+  "name": "Aurora MainNet",
   "chain": "NEAR",
   "network": "mainnet",
   "rpc": [
+    "https://rpc.mainnet.aurora.dev:8545"
   ],
-  "faucets": [
-  ],
+  "faucets": [],
   "nativeCurrency": {
-    "name": "NEAR",
-    "symbol": "NEAR",
-    "decimals": 24
+    "name": "Ether",
+    "symbol": "ETH",
+    "decimals": 18
   },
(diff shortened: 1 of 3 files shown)


ligi:
Thanks - I am not really sure about the new native currency definitions. It's not ETH right? IMHO they should also be distinct

artob:
@ligi It's bridged ETH. That is, all ETH in circulation on the Aurora network was transferred over from Ethereum via the Aurora Bridge. That locks up the corresponding backing ETH on Ethereum. So, effectively, the base token on Aurora is a 1:1-redeemable IOU for ETH on the Ethereum mainnet.

ligi:
Even on the TestNets? Still think it should be different to original ether - maybe bETH for (bridged Ether?)

artob:
@ligi Yes, our testnet is bridged to Ropsten, allowing all the same functionality. (I note that the base token on Goerli, Rinkeby, and Ropsten is also listed as just ETH in this registry.)

To be clear, we aren't planning on prefixing or suffixing the base token name with anything ourselves. We've considered the question at length, and it will be just ETH in all user interfaces.

However, for the purposes of this registry, I hear you and I've [pushed a change to update it to aETH](https://github.com/ethereum-lists/chains/pull/259/commits/ed5bf3892cfabecd8a837af7f23e684c90163101), for Aurora ETH, if that's more to your liking.

ligi:
It's not just me - (very) recently started a poll and a majority is for not allowing bridged chains use ETH as the symbol:
https://twitter.com/mr_ligi/status/1393135790948036610

So looks good now - thanks for the change and will merge now. That said you should think if you do not want to distinguish between your main-net and test-net.

ligi (approved):

artob:
Thanks, @ligi. I'll raise the question internally and come back with a follow-up PR if need be.

Hlsinclair:
I attached pics. the recieving wallet is Thunder Cores special Wallet. My
send never went to any bridge, it's in that wallet which is Also mine. They
want money and won't give me or help me. I could really use my money back.
I can't figure out or understand how to reverse the transaction, or go
steal it because there is no way to send it back. It's their wallet...they
want more money and I can't and won't..I'm a Vet.

PR #265, opened 2021-05-20
change the shortname and chain

vincentbdb:
suddenly notice the short name was used by another coin
tks


ligi:
which other chain uses he shortName? There should be no duplicate and I do not see it.

vincentbdb:
> which other chain uses he shortName? There should be no duplicate and I do not see it.

not another chain, I mean there is one ERC20 token in ETH have the name GTC. so we use GT simple and clear. tks

PR #267, opened 2021-05-20
Add orbitchain

cinnaminty-d:


ligi (changes requested):
The explorer must adhere to a standard

drop-out-dev:
Is this the case for all new pull requests? The explorer doesn't quite adhere to EIP3091 yet.

ligi:
since the explorer field is introduced I check for the standard.

PR #277, opened 2021-05-28
Add Arbitrum mainnet

fredlacs:


_data/chains/eip155-42161.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "Arbitrum One",
+  "chainId": 42161,
+  "shortName": "arb1",
+  "chain": "ETH",
+  "network": "mainnet",
+  "networkId": 42161,
+  "nativeCurrency": {
+    "name": "Ether",
+    "symbol": "ETH",
```
ligi, line 10: The symbol cannot be ETH. We asked the community and only Ethereum L1 should be able to have the ETH symbol. Maybe use AETH?

ligi (changes requested):

_data/chains/eip155-42161.json
gsteele178: Thank you

gsteele178 (approved):

_data/chains/eip155-42161.json
fredlacs: I see, I'm curious as to what the rationale behind this is.
Is there an ethereum magicians thread or something of the kind I could read up on?

ligi: I ran this poll on twitter as it is more about users than about devs: https://twitter.com/mr_ligi/status/1393135790948036610
there are some reasons in the replies.

gsteele178: There is no rational behind any of this.... That's the beauty of it

ligi: *?*

gsteele178:
The eth in etherscan, I try to withdraw it through meta mask 9 times out of
10 fails...

ligi:
@fredlacs can you change the symbol so this can be merged?

_data/chains/eip155-42161.json
```diff
@@ -0,0 +1,19 @@
...
+    "decimals": 18
+  },
+  "rpc": [
+    "https://arb1.arbitrum.io/rpc",
+    "wss://arb1.arbitrum.io/ws"
+  ],
+  "faucets": [],
```
credfeto, line 17: Can you add the block explorer to this too: https://explorer.arbitrum.io

_data/chains/eip155-42161.json
mccordnate: @ligi I'm not sure Arbitrum necessarily fits under the same criteria here. For a typical bridge, it definitely makes sense to restrict it to wrapped representations when denominating in ETH. However, given that Arbitrum is an L2 roll-up solution it's fair to say this is the same ETH as what you have on Ethereum. The reason being that you can never lose the ETH in your Arbitrum wallet. Even if the sequencer goes down, transactions can be posted to the L1 contract by other parties. This allows withdrawals to always be possible.

ligi: What is if there is a bug in the Arbitrum (or any L2) contracts? I still think there is a difference between ETH and L2-ETH - but I see your point that L2-ETH is closer to ETH than other derivatives. Ideas welcome on how to convey this to users - but IMHO we cannot call it ETH.

rahul-kothari: Chain-list uses ARETH for arbitrum

ligi: What do you refer to as Chain-list?

rahul-kothari:
https://chainlist.org/

*Regards,*
*Rahul Kothari*
*www.linkedin.com/in/iamrahulkothari
<http://www.linkedin.com/in/iamrahulkothari>*
*www.twitter.com/iamRahul20x* <http://www.twitter.com/iamRahul20x>

ligi:
chainlist.org uses the data from this repo ;-)
And ARETH is the symbol for the Rinkeby Testnet of Arbitrum

ligi (approved):

fredlacs:
Sorry for the slow response time @ligi 
I updated the currency symbol to `AETH` as you suggested, also went ahead and updated explorers/rpc support

PR #292, opened 2021-06-11
KCC MAINNET

lingyun92:


ligi (changes requested):
short name must be unique - but found `kcc` more than once

lingyun92:
> short name must be unique - but found `kcc` more than once
Thanks for your review, I updated the shortName of the testnet.

PR #294, opened 2021-06-11
Add Proton TestNet(eip155-110)

snrntpa2z:


ligi (changes requested):
The explorer does not seem to be EIP3091 compatible.

ligi (approved):

PR #307, opened 2021-06-21
Add ArbRinkeby

FabijanC:
native currency?

source: https://twitter.com/arbitrum/status/1403102638699433984

### DIFF
--- a/_data/chains/eip155-421611.json
+++ b/_data/chains/eip155-421611.json
@@ -6,8 +6,8 @@
   "network": "rinkeby",
   "networkId": 421611,
   "nativeCurrency": {
-    "name": "Arbitrum Ether",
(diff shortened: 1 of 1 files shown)


ligi (changes requested):
I think the currency should be Arbitrum Rinkeby Ether (ARETH)

ligi (approved):

vmichalik:
@ligi it doesn't seem like the Arbitrum team are in alignment over this naming for the native currency, see their documentation [here](https://developer.offchainlabs.com/docs/public_chains)

image

I'll make a PR to change it to RinkebyETH

ligi:
Dam - I actually think this is worse - but if it's the official one - yea. Naming is hard.

PR #308, opened 2021-06-23
Add hoo smart chain mainnet

HooGroup:
Add hoo smart chain mainnet


ligi (changes requested):
short name must be unique - but found `hoosmartchain` more than once

PR #317, opened 2021-07-02
add quarkchain chainids

ping-ke:
add quarkchain chainids
* mainnet: https://mainnet.quarkchain.io/
* devnet: https://devnet.quarkchain.io/

### DIFF
--- a/_data/chains/eip155-100000.json
+++ b/_data/chains/eip155-100000.json
@@ -1,0 +1,19 @@
+{
+  "name": "QuarkChain Mainnet Root",
+  "chain": "QuarkChain",
+  "network": "mainnet",
+  "rpc": [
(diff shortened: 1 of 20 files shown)


ping-ke:
so I need to add a parent field, right? and parent:chain will be the chain name of the root? is the root chain also need to add a parent field?
 "parent" : { 
   "chain": "CAIP-2 of parent",
   "type": "shard"
 }

ligi:
Yea - I will add support for the parent field later today. Then you can add it as you said. But only for the shards - not for the root - I am thinking the root has no parent - otherwise ti would be a weird root ;-)

ligi:
just added the parent field support - if you add them to your shards I will merge - thanks!

PR #323, opened 2021-07-05
HECO chain (128): Change infoURL to explorer URL. Fix infoURL.

Wetitpig:
The infoURL seems to be an explorer URL rather than a URL for the information of the chain. <https://www.hecochain.com> was added back as it provides information of the chain.


ligi:
thanks!

PR #331, opened 2021-07-07
add ontology chainid

lucas7788:


ligi (changes requested):
Unfortunately it seems that the explorer is not EIP3091 compatible

e.g. https://explorer.ont.io/testnet/block/16089828 does not work

it is: https://explorer.ont.io/block/16089828/testnet - but that violates the EIP

noncash15:
#331

noncash15 (commented):
_data/chains/eip155-5851.json

lucas7788:
> https://explorer.ont.io/testnet/block/16089828

we have fix this problem, please merge the pull request as soon as possible. Thank you.

PR #335, opened 2021-07-14
Add chain ID for the cheapETH blockchain

geohot:
cheapETH is alive and well and using chain ID 777!

Learn more at https://cheapeth.org, perhaps more legit than the Ethermint Testnet?

### DIFF
--- a/_data/chains/eip155-777.json
+++ b/_data/chains/eip155-777.json
@@ -1,18 +1,18 @@
 {
-  "name": "Ethermint Testnet",
-  "chain": "ETHERMINT",
-  "network": "testnet",
+  "name": "cheapETH",
+  "chain": "cheapETH",
+  "network": "cheapnet",
   "rpc": [
-    "http://54.210.246.165:8545"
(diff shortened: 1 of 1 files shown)


ligi:
In a case of a collision the oldest chain wins and it seems yours is younger

ligi:
OK - just wanted to check the timestamps of the genesis block - and it seems their RPC is down - so there is a chance actually ;-)

fedekunze (changes requested):
We will still use 777 for our testnet chain-id, we stopped it because we were doing breaking changes and also because the maintainers of the project changed. We will relaunch the testnet in a few weeks

ligi:
Thanks for the information! Will 'relaunch' mean a new genesis?

ligi:
OK merged this now as the  Ethermint Testnet is inactive. Once it comes back (with the same genesis) - I might revert. But for users I think this is the best way now.

geohot:
Thanks!

PR #343, opened 2021-07-19
add plian

terry108:
Add Plian mainnet and testnet

### DIFF
--- a/_data/chains/eip155-10067275.json
+++ b/_data/chains/eip155-10067275.json
@@ -20,5 +20,9 @@
     "name": "piscan",
     "url": "https://testnet.plian.org/child_test",
     "standard": "EIP3091"
-  }]
+  }],
+  "parent": {
+    "chain": "eip155-16658437",
+    "type": "L2"
+  }
 }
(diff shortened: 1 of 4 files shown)


ligi (changes requested):
I cannot open the explorer URLs - please remove when they are not active yet or make this PR a draft when I should wait until they are active.
Also "subchain" maybe we want to use the parent field here? Are these subchains L2's to the main-chains?

terry108:
> I cannot open the explorer URLs - please remove when they are not active yet or make this PR a draft when I should wait until they are active.
> Also "subchain" maybe we want to use the parent field here? Are these subchains L2's to the main-chains?

The explorer URLs is working now.  Please check it again.
Yes, subchains are L2's to the main-chains.

daishiydmaguchi:
null

ligi (changes requested):
can you please add a parent field like so:
```json
 "parent": {
    "chain": "eip155-XXX",
    "type": "L2"
  }
```

where eip155-XXX is the parent chain?

ligi:
we just need to make the CI happy to merge - and CI complains about the short name not being unique

terry108:
> we just need to make the CI happy to merge - and CI complains about the short name not being unique

Done, please have a try again. Txs~

ligi:
hm - I do not really understand your change - but the shortname must be unique - means each chain has a different shortName - currently they all have the same ..

PR #345, opened 2021-07-20
Moonriver update, add Moonshadow

purestaketdb:
Update Moonriver chain currency and add RPC URLs

Add occasionally used Moonshadow chain ID


ligi:
thanks!

PR #354, opened 2021-07-28
Update the explorer URLs for the Theta Mainnet and testnets

jieyilong:


ligi (changes requested):
the explorers do not seem EIP3091 compatible - e.g.

https://testnet-explorer.thetatoken.org/block/11665537

does not work - but:


https://testnet-explorer.thetatoken.org/blocks/11665537 work

jieyilong:
> the explorers do not seem EIP3091 compatible - e.g.
> 
> https://testnet-explorer.thetatoken.org/block/11665537
> 
> does not work - but:
> 
> https://testnet-explorer.thetatoken.org/blocks/11665537 work

Thanks for pointing that out. We have fixed the `block/` endpoint. Please let us know if you any further feedback.

PR #367, opened 2021-08-10
Add Proton Mainnet(eip155-49)

snrntpa2z:


ligi:
wondering why block 2 does not exist: https://www.protonscan.io/block/2

ligi (changes requested):
seems to need a redirect tokens -> token to adhere to EIP3091
this works:
https://www.protonscan.io/tokens/XPR-proton-eosio.token

this does not:
https://www.protonscan.io/token/XPR-proton-eosio.token

PR #368, opened 2021-08-11
support https

lucas7788:


ligi (changes requested):
maybe we should remove them completely as they do not seem to work on both protocols.

lucas7788:
> maybe we should remove them completely as they do not seem to work on both protocols.

We have fixed the problem that the test network cannot be used. The main network will also be available after we go online.

ligi:
and can you provide the endpoints via https? Providing them via http outside the dappnode vpn is not really good style ..

OhArthits (approved):

lucas7788:
> and can you provide the endpoints via https? Providing them via http outside the dappnode vpn is not really good style ..

thanks for your advice, wo will support https, but we need to use the new port. please take a look again.

lucas7788:
> and can you provide the endpoints via https? Providing them via http outside the dappnode vpn is not really good style ..

we have support https link,  please have a look again. thanks

lucas7788:
> and can you provide the endpoints via https? Providing them via http outside the dappnode vpn is not really good style ..

we have support https, could you please take a look.

PR #378, opened 2021-08-18
Add BitkubChain mainnet and testnet

isman-usoh:
- https://www.bitkubchain.com/
- add BitkubChain mainnet `chainId=96`
- add BitkubChain testnet `chainId=25925`

### DIFF
--- a/_data/chains/eip155-25925.json
+++ b/_data/chains/eip155-25925.json
@@ -21,8 +21,7 @@
   "explorers": [
     {
       "name": "BKCScan Testnet",
-      "url": "https://testnet.bkcscan.com",
-      "standard": "EIP3091"
+      "url": "https://testnet.bkcscan.com"
     }
   ]
 }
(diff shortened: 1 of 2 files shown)


ligi (changes requested):
The explorers do not seem to be EIP3091 compatible - e.g.
https://testnet.bkcscan.com/block/2314228
fails 
and only 
https://testnet.bkcscan.com/blocks/2314228 works

isman-usoh:
> The explorers do not seem to be EIP3091 compatible - e.g.
> https://testnet.bkcscan.com/block/2314228
> fails
> and only
> https://testnet.bkcscan.com/blocks/2314228 works

I fixed it

ligi (changes requested):
You need to remove the whole explorer section - we do not accept explorers that do not adhere to any standard

ligi (changes requested):
The short name must be uniq

PR #381, opened 2021-08-23
add the metadot chain info

bengbengle:


ligi (changes requested):
name and shortName must be unique

PR #382, opened 2021-08-24
Adding Scallop Testnet to the List

pawan-scallop:
Dear All,

We are launching our own chain at Scallop for a specific usecase. You can find more details about us at www.scallopx.com.

Kindly do the needful.

regards
Pawan Pandey
CTO- Scallop Group


ligi (changes requested):
You checked in a lot of files that really should not be in the PR - please remove them

ligi (changes requested):
also the files in .gradle should not be added by this PR - PS: checking out the repo and deleting all of them would be much faster than deleting them one by one via one commit

PR #389, opened 2021-08-31
Add information for Arbitrum on xDai

akolotov:
Consider to add the information about Arbitrum on xDai rollup which is associated with the chain id 200.

### DIFF
--- a/_data/chains/eip155-200.json
+++ b/_data/chains/eip155-200.json
@@ -1,0 +1,21 @@
+{
+  "name": "Arbitrum on xDai",
+  "chain": "AOX",
+  "network": "xdai",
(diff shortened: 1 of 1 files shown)


ligi (changes requested):
so it is a L2 on top of xDAI? If so can you add a parent field then?
like so:
https://github.com/ethereum-lists/chains/blob/c8477c74105528d9b829fbe53bc8c71293fd6dd5/src/test/resources/test_chains/valid/withparent/eip155-2.json#L19

akolotov:
> thanks - merged it - but maybe you can add the parent field later if it is a L2 of xDAI

yes, it is. I will add the parent -- thanks for the example above.

PR #392, opened 2021-08-31
Support syscoin

tomatoskittles:
We would like to add Syscoin and Syscoin Testnet. Please let me know if there is anything else you would like me to do.

### DIFF
--- a/_data/chains/eip155-57.json
+++ b/_data/chains/eip155-57.json
@@ -7,7 +7,7 @@
     "https://web3.syscoin.org",
     "wss://web3.syscoin.org"
   ],
-  "faucets": ["https://faucet.syscoin.org"]
+  "faucets": ["https://faucet.syscoin.org"],
   "nativeCurrency": {
     "name": "Syscoin",
     "symbol": "SYS",
@@ -16,9 +16,5 @@
   "infoURL": "https://www.syscoin.org",
   "shortName": "sys",
   "chainId": 57,
-  "networkId": 57,
-  "explorers": [{
-    "name": "syscoin explorer",
-    "url": "https://nevm.syscoin.org"
-  }]
+  "networkId": 57
 }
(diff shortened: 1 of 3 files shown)


_data/chains/eip155-57.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "Syscoin Mainnet",
+  "chain": "SYS",
+  "network": "mainnet",
+  "rpc": [
+    "https://nevm.syscoin.org/api/eth-rpc",
+    "https://web3.syscoin.org",
+    "wss://web3.syscoin.org"
+  ],
+  "faucets": ["https://faucet.syscoin.org"]
```
ligi, line 10:
```suggestion
  "faucets": ["https://faucet.syscoin.org"],
```

_data/chains/eip155-5700.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "Syscoin Tanenbaum Testnet",
+  "chain": "SYS",
+  "network": "testnet",
+  "rpc": [
+    "https://tanenbaum.io/api/eth-rpc"
+  ],
+  "faucets": ["https://faucet.tanenbaum.io"]
```
ligi, line 8:
```suggestion
  "faucets": ["https://faucet.tanenbaum.io"],
```

ligi (changes requested):
there are 2 commas missing - also the explorers need the `standard` field

tomatoskittles:
> there are 2 commas missing - also the explorers need the `standard` field

Thanks for the review.

I have fixed the comma related syntax error, and I have also removed the explorers. We are using Blockscout, which I don't believe is EIP3091 compliant. xDai also uses Blockscout, and their entry doesn't have explorers either, so I figured this was a good approach.

Hopefully it is good, now.

ligi (approved):

ligi (changes requested):
short name must be unique - but found `sys` more than once

tomatoskittles:
> short name must be unique - but found `sys` more than once

I've changed the testnet shortname, to honour the uniqueness requirement.

PR #396, opened 2021-09-03
Updates to Match MM

purestaketdb:
Removing unnecessary parent relay chain name from "name".


ligi:
are these L2s on top of other chains? Maybe you want to add the new "parent" field then

purestaketdb:
Thanks @ligi , they are L1's on top of Substrate's L0 (Polkadot, Kusama, Westend, etc.).

ligi:
oh - L0 - interesting ;-)
but we can only reference layers below that are in this lists - so this will not work I think - thanks anyay!

PR #397, opened 2021-09-03
Add OneLedger mainnet and testnet

DeV1doR:


ligi (approved):

ligi (changes requested):
short name must be unique - but found `olt` more than once

DeV1doR:
@ligi fixed

PR #410, opened 2021-09-13
add chain configs for pulsechain main and test net

3commascapital:


ligi (changes requested):
ShortNameMustBeUnique: short name must be unique - but found `pls` more than once

3commascapital:
ah! thanks for the comment. i'll converse w/ team and update

pulseoglank (changes requested):
You have the wrong Chain ID. The official Chain ID: 940

3commascapital:
@pulseoglank this is for testnet which is accurate

PR #423, opened 2021-09-16
Create 以太坊2lL

jrvv123:


ligi (changes requested):
can you elaborate? Makes no sense to me tbh.

PR #430, opened 2021-09-22
Fixed Celo explorer URL

Deepcryptodive:
Added a Celo explorer URL.

Previously, Metamask used the infoURL which forwards users to a 'Page Not Found' page, instead of a block explorer.

image


ligi (changes requested):
The explorer does not seem to be EIP3091 compatible. 
just tried:
https://explorer.celo.org/block/8957819
which does not work
https://explorer.celo.org/blocks/8957819
works - there needs to be e.g. an redirect

PR #455, opened 2021-10-05
Create eip155-1444.json

Alireza-Zojaji:
Add SUR Blockchain Network


ligi (changes requested):
The ChainID in the filename does not match the chainID in the content

ligi:
can this now be closed after #457 ?

PR #463, opened 2021-10-07
add missing explorers, faucet and fixed some ids

porkybalboa:


ligi:
Thanks for the PR! There is useful stuff in there - just a bit torn about 2 things:
 * the faucet has rekt.al in the top - which is really weird and maybe not nice for some users in a language that I speak unfortunately
 * some explorers like the celo one do not respect EIP3091 - wondering now if it is better to make the standard field optional. On the other side we had some cases where people adopted EIP3091 to get an explorer in the data.

porkybalboa:
>     * some explorers like the celo one do not respect EIP3091 - wondering now if it is better to make the standard field optional. On the other side we had some cases where people adopted EIP3091 to get an explorer in the data.

i noticed the latest version of blockscout respect the uri route schema from the EIP3091 along with their old routes in parallel. i guess some customized forks based of older blockscouts probably miss this change. but along my multichain adventures i hardly came across these.
maybe leave the stand-field in, urging devs to adopt the EIP. by itself patching in these routes should no problem for someone who is able to customize a blackexplorer. once the users complain that their explorer links do not work expectedly they will be quick to adopt, i guess.

PR #484, opened 2021-10-13
Add Velas Mainnet

roman-tik:


roman-tik:
> Required value 'faucets' missing - can be an empty array

done, thx

roman-tik:
@ligi

PR #488, opened 2021-10-13
Removing Syscoin mainnet faucet, and adding explorers.

tomatoskittles:
Updating information for Syscoin:

Removed the mainnet faucet.

Added explorers to both.


ligi:
yes

tomatoskittles:
> yes

Ok, I have removed the addition of the explorer, and I'll add it back in later, when it is up.

PR #518, opened 2021-10-28
Add Ambros Chain Mainnet and Testnet

anhcuong:
Add Ambros Chain Mainnet and Testnet


anhcuong:
@ligi i have updated the shortname. Ambros is a private chain of bcmhunt blockchain game

PR #520, opened 2021-10-29
add bittorrent chain testnet

zhaohong:


bertux:
doesn't seem to match anymore https://doc.bt.io/v1/doc/testnet.html#bttc-testnet-2

PR #532, opened 2021-11-02
add Thinkium Network

ThinkiumChain:
add Thinkium Network

### DIFF
--- a/_data/chains/eip155-60000.json
+++ b/_data/chains/eip155-60000.json
@@ -1,0 +1,25 @@
+{
+    "name": "Thinkium Testnet Chain 0",
+    "chain": "Thinkium",
+    "network": "testnet",
+    "rpc": [
+        "https://test.thinkiumrpc.net/"
+    ],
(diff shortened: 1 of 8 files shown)


ligi (changes requested):
The shortNames must be unique
Also the explorers cannot fulfill EIP3091 as they are the same URLs for different networks

ligi:
@matejcik can chains with different chainID have the same slip44?

matejcik:
multiple chains of the same network, yes.

but this is reusing SLIP44 ID for Matic, which it should not be doing

ilovechain:
hi, I‘v changed already according to rules. is it correct now ?

ThinkiumChain:
hi, I‘v changed already according to the rules. is it correct now ?

ThinkiumChain:
hi, I‘v changed already according to the rules. is it correct now ?  @matejcik @ligi

ligi (approved):

PR #538, opened 2021-11-04
add Meter Testnet, update Meter Mainnet

simonzg:


ligi (changes requested):
short name must be unique - but found `Meter` more than once

simonzg:
@ligi I've made the changes for shortName, please review again. Thanks

PR #541, opened 2021-11-05
add shibachain

shibachain:


ligi (changes requested):
The block explorer does not seem to respect EIP3091 - please change to 'none'

shibachain:
Changed. Thank you.

PR #557, opened 2021-11-11
Adding NEXT Smart Chain

dcnl1980:


ligi (approved):

ligi:
This chain will likely be removed - there seems to be an older one with the same chainID:

https://github.com/ethereum-lists/chains/pull/1629

dcnl1980:
Next smart chain will not be removed and is up and running. Please refuse the request.

ligi:
seems to be younger than the other one though - so as for https://github.com/ethereum-lists/chains#collision-management
it will be removed

PR #566, opened 2021-11-16
Fix explorer address

henry-hz:


ligi (changes requested):
The explorer should go in the explorer section and not in infoURL

henry-hz:
yep, but when I click the "see transaction" in the Metamask, it sends to
this addres

CHANDRAMA2:
#592

PR #578, opened 2021-11-21
Create eip155-281121 Social Smart Chain Network

thedarkknight197:


thedarkknight197:
Anyone could check? not my file give error.

thedarkknight197:
The error is in: _data/chains/eip155-64.json
My file is _data/chains/eip155-281121.json

PR #591, opened 2021-11-23
Add Metis L2 testnet and parent

pavelsinelnikov:


ligi (changes requested):
the shortName must be unique - also surprised by the rinkeby testenet - shouldn't it have rinkeby as parent then?

pavelsinelnikov:
My bad, I have hopefully corrected this error in the latest commit

PR #598, opened 2021-11-26
add Gather chains

scream4ik:
Create entries for Gather Network (mainnet, testnet, devnet)

### DIFF
--- a/_data/chains/eip155-192837465.json
+++ b/_data/chains/eip155-192837465.json
@@ -1,0 +1,23 @@
+{
+  "name": "Gather Mainnet Network",
+  "chain": "GTH",
+  "network": "mainnet",
+  "rpc": [
+    "https://mainnet.gather.network"
+  ],
+  "faucets": [],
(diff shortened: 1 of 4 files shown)


_data/chains/eip155-192837465.json
```diff
@@ -0,0 +1,23 @@
...
+    "https://mainnet.gather.network"
+  ],
+  "faucets": [],
+  "nativeCurrency": {
+    "name": "Gather",
+    "symbol": "GTH",
+    "decimals": 18
+  },
+  "infoURL": "https://gather.network",
+  "shortName": "Gather Mainnet",
```
rmeissner, line 15: Is this really a shortname?

biafra23: Please use `gth`. 
As far as I understand, this will be used as the prefix to all addresses on this chain.

scream4ik: fixed

PR #681, opened 2021-12-16
create eip155-2025.json &&  eip155-9527.json

FlyingLittleBear:
we create new chain rangers which use the chain id 2025,and testnet use chain id 9527.


ligi (changes requested):
short name must be unique - but found `rpg` more than once

PR #696, opened 2021-12-21
Update Gnosis chain

GimluCom:
Updated Gnosis chain, formerly xdai chain.


ligi (commented):
wondering if we should put (formerly xDAI) into brackets here for the name - think a lot of people will still search for xDAI and have no results otherwise - What do you think?

GimluCom:
Looks like it will appear considering the native currency still is xDAI, however it is probably a good idea to add it for clarification.

PR #698, opened 2021-12-23
New chain id

kawisin:
New chainid


kawisin (commented):
this for testing system chain network. want to develop

PR #708, opened 2021-12-25
Add Emerald Paratime networks.

0xatm:
Add Emerald Paratime Mainnet and Testnet.


ligi (changes requested):
the explorer seems to obey eip-3091 - so you can use it in the standard field - or is there any reason you used `none`?

0xatm:
> the explorer seems to obey eip-3091 - so you can use it in the standard field - or is there any reason you used `none`?

Great spot, I've updated them all, thanks for the review.

PR #743, opened 2022-01-07
Create eip155-45528

bradleyjw:
Add Everest EverChain


ligi (changes requested):
I cannot reach the explorer - also the RPC is wrong

bradleyjw:
Thank you - we are reviewing and will resubmit when ready.

PR #752, opened 2022-01-10
Create eip155-47805.json and rei.json

samlior:
https://rei.network/


ligi (changes requested):
The explorer does not seem to be EIP3091 compatible - so you need to set standard to "none" or even better fix it

samlior:
Sorry for that, I've changed it to "none"

PR #754, opened 2022-01-10
Update coston testnet details

RareData:


ligi:
maybe you would want to improve the faucet so you can pass a address - see e.g. chain 5 as an example. You can also send funds to https://fauceth.komputing.org/admin and I add the chain

RareData:
That'd be appreciated. 1,000,000 CFLR has been sent to your address.

ligi:
FaucETH now serves this testnet

PR #790, opened 2022-01-21
Add UBC

QDanlSc:


ligi (changes requested):
shortNames must be unique - also the testnet chainId/networkID is wrong

QDanlSc:
> shortNames must be unique - also the testnet chainId/networkID is wrong

has changed

ligi (changes requested):
name must be unique - but found `UB Smart Chain` more than once

ligi (approved):

PR #796, opened 2022-01-22
Adding AVO Network

blockchainreg:


ligi (changes requested):
I cannot reach https://scan.avo.network to check if it is EIP3091 compatible

Kalakaarboyz1 (approved):

PR #809, opened 2022-01-27
add garizon chain

portuu3:


ligi (changes requested):
The shortnames must be unique - also maybe add a parent field

PR #826, opened 2022-01-31
Add eCredits mainnet & testnet

ecsdev2:


ligi (changes requested):
filename extension must be json

ecsdev2:
> filename extension must be json

Sorry, silly mistake

ligi (changes requested):
short name must be unique - but found `ecs` more than once

PR #832, opened 2022-02-04
Add Palm RPC endpoints

mbaxter:
### Description

Add RPC endpoints for the Palm network.


ptescher:
@mbaxter FYI every other RPC endpoint uses `${INFURA_API_KEY}` but you added it as `{INFURA_API_KEY}`

mbaxter:
> @mbaxter FYI every other RPC endpoint uses `${INFURA_API_KEY}` but you added it as `{INFURA_API_KEY}`

@ptescher - nice catch and thanks for the heads up, will push up a fix!

PR #843, opened 2022-02-08
Update Optimism chain info

bdresser:
Following up on conversation in https://github.com/ethereum-lists/chains/pull/203 – updating Optimism's currency symbol to keep in line with emerging standard across wallets & apps. 

cc @ligi

### DIFF
--- a/_data/chains/eip155-10.json
+++ b/_data/chains/eip155-10.json
@@ -5,11 +5,11 @@
   "faucets": [],
   "nativeCurrency": {
     "name": "Ether",
-    "symbol": "OETH",
+    "symbol": "ETH",
     "decimals": 18
   },
   "infoURL": "https://optimism.io",
-  "shortName": "oeth",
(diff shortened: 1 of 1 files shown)


_data/chains/eip155-10.json
```diff
@@ -5,11 +5,11 @@
   "faucets": [],
   "nativeCurrency": {
     "name": "Ether",
-    "symbol": "OETH",
+    "symbol": "ETH",
     "decimals": 18
   },
   "infoURL": "https://optimism.io",
-  "shortName": "oeth",
+  "shortName": "eth",
```
ligi, line 12:
```suggestion
  "shortName": "oeth",
```

ligi:
OK after having some more discussion around it I decided to merge it now - but except the short name change - the short name MUST be unique - this is a hard requirement e.g. for TREZOR and eip-3770 - no chance there can be a collision

PR #853, opened 2022-02-09
Add Nebula Testnet

nova-network-inc:


ligi (changes requested):
Same as the other one - you need to rebase to master so that the icon is included

nova-network-inc:
> Same as the other one - you need to rebase to master so that the icon is included

Hi, I'm quite lost here, do I commit again for both networks, with separate icon directories?

PR #875, opened 2022-02-13
Add HTTPS RPCs and Explorer

nova-network-inc:
Added secure RPCs and Explorer.


nova-network-inc:
> do you have an tx hash and address so I can check EIP3091 compatibility?

Sure, here https://explorer.novanetwork.io/tx/0xef31297c08dba111f00577c79bb2945da02dbadc64ff06289655e0915d0d23ea

PR #908, opened 2022-02-21
teleport testnet

KaelTop:
Add [Teleport](https://teleport.network/) Testnet Chain ID **8001**

Docs: https://chain-docs.teleport.network

Get Chain ID:

```bash
curl -X POST -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"net_version","params":[],"id":67}' https://evm-rpc.testnet.teleport.network | jq .result
```


ligi (changes requested):
The icon musb e on ipfs://

PR #917, opened 2022-02-22
Update eip155-9000.json

litvinsky:
Hi from Evmos team! Can you please update our settings? We started up a new testnet!


ligi:
thanks! PS: if you send some testnet ether to https://fauceth.komputing.org/address then I can add you there - would have e.g. the feature that e.g. wallets can direclty pass an address & this way make it easier for users.

PR #925, opened 2022-02-24
Update Avalanche C Chain values

FrederikBolding:
Updated Avalanche values to make it clear that the chain is the Avalanche C Chain.

### DIFF
--- a/_data/chains/eip155-43114.json
+++ b/_data/chains/eip155-43114.json
@@ -1,5 +1,5 @@
 {
-  "name": "Avalanche Mainnet",
+  "name": "Avalanche C-Chain",
   "chain": "AVAX",
   "rpc": [
     "https://api.avax.network/ext/bc/C/rpc"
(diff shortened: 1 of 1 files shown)


ligi (changes requested):
Thanks! But the chainID in the filename must match the field

FrederikBolding:
@ligi Not sure if you also want `chain` to be corrected to `AVAXC` or something like that? 

There has been some confusion from our side on the paths and as such we had added the wrong `slip44` value for the C-Chain.

FrederikBolding:
> Thanks! But the chainID in the filename must match the field

ChainID does match! The only thing changed other than the name is the `slip44` value, which is currently wrong for the C-Chain!

ligi (approved):

ligi:
Sorry - you are right - was looking at it between 2 other things and thought the number change is the chainID

PR #946, opened 2022-03-06
Replace chainlist.org with evmchainlist.org

phx9:
Because chain list will be unavailable at 4.3


ligi (changes requested):
Please do not replace the site - so far it is still up and might live on - see: https://github.com/antonnell/networklist-org/issues/58
You can add yours (on the bottom) - but please do not remove the current one.

PR #959, opened 2022-03-11
Create Hupayx Network 2023, 22023

ej-ri:
Add chain id 2023, 22023


ligi (changes requested):
short name must be unique - but found `hpx` more than once

also please support our grant: https://gitcoin.co/grants/2974/ethereum-lists

suriyawut2344 (approved):

PR #960, opened 2022-03-11
Update Aurora explorer and symbol.

paouvrard:
Aurora networks use ETH as the native currency to pay for gas so adding networks to MetaMask gave a warning because aETH is currently registered.
Update new explorer URLs.


ligi (approved):
LGTM.

Please also support our grant: https://gitcoin.co/grants/2974/ethereum-lists

PR #962, opened 2022-03-11
Update eip155-108.json

ErazerTriplet3:
Changing the info url allows metamask to trace transactions when clicked.
currently in metamask if you trace the transaction when clicked it just goes to the main thundercore page instead of the scanner


ligi (changes requested):
you can add a explorer section - but the infoURL should not be an explorer

please also think about supporting our grant: https://gitcoin.co/grants/2974/ethereum-lists

PR #964, opened 2022-03-11
Update eip155-18.json

ErazerTriplet3:
Change the info url to the thundercore scanner.
This means when in metamask and a user clicks the trace transaction button it takes them to the scanning link of the transaction.
Currently leaving it as the thundercore.com website means clicking the scan in metamask just takes the user to thundercores website

### DIFF
--- a/_data/chains/eip155-18.json
+++ b/_data/chains/eip155-18.json
@@ -12,7 +12,7 @@
     "symbol": "TST",
     "decimals": 18
   },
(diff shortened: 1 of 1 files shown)


ligi (changes requested):
you can add a explorer section - but the infoURL should not be an explorer

please also think about supporting our grant: https://gitcoin.co/grants/2974/ethereum-lists

Lucid369 (approved):

PR #980, opened 2022-03-15
Adding Frost Testnet

icedevportal:


Mosaleh0114:
Ok what do i do now

ligi:
fix the problem that the IDs do not match ..

PR #982, opened 2022-03-15
Adds v3 and v2b

3commascapital:


ligi (changes requested):
name must be unique - but found `PulseChain Testnet` more than once

3commascapital:
thanks @ligi . just updated. do the tests need to be restarted? seems like they never finished

PR #983, opened 2022-03-16
Fix incorrect Venidium testnet path

HaiDang666:


ligi (changes requested):
You must have a `faucets` field - this can be an empty array if you have no faucet

ligi (approved):

ligi:
Please also consider our grant: https://gitcoin.co/grants/2974/ethereum-lists

PR #1002, opened 2022-03-25
Add DFK Chain

MrFaun:
This PR adds the details for the DFK Chain to chainlist. DFK Chain is an evm chain launched on avalanche subnets.


ligi (changes requested):
short name must be unique - but found `DFK` more than once

Gunawanwahyudi21:
Good

MrFaun:
Changed test shortname

PR #1010, opened 2022-03-29
Add Nahmii mainnet and testnet

JHolba:


ligi (changes requested):
explorer seems to be not EIP3091 compatible - please change to `none`

JHolba:
I will upgrade our explorer to support EIP3091. It should be live on mainnet and testnet early next week. I will update when it is done. Sorry for the delay in answering.

JHolba:
Support for EIP3091 is now available on both testnet and mainnet. Is any other change needed?

PR #1011, opened 2022-03-29
add fomochain

HeroMetaverse:


ligi (changes requested):
short name cannot be empty

PR #1040, opened 2022-04-02
Change PlatON devnet's chainid from 210309 to 2203181

benbaley:


ligi (changes requested):
You cannot just change a chainID - this would open up the door to replay attacks. You can add a new chain and deprecate the old one.

ligi:
OK it loaded after a long while

PR #1055, opened 2022-04-06
Update cennznet

aidan-starke:
- update chain/network id to 21337
 - add width & height properties to cennz icon

### DIFF
--- a/_data/chains/eip155-1337.json
+++ b/_data/chains/eip155-1337.json
@@ -1,10 +1,7 @@
 {
-  "name": "CENNZnet Azalea",
+  "name": "CENNZnet old",
   "chain": "CENNZnet",
-  "network": "azalea",
-  "rpc": [
-    "https://cennznet.unfrastructure.io/public"
-  ],
+  "rpc": [],
   "faucets": [],
   "nativeCurrency": {
     "name": "CPAY",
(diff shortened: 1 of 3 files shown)


ligi (changes requested):
You cannot just change a chainID - you need to deprecate the old chain and add a new one - otherwise this opens doors to replay attacks
Also the size of the SVG is weird - and vector assets do not need a size

ligi:
Also:

```
✓ ~ curl -X POST -H  "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_chainId","params":[],"id":1}' https://cennznet.unfrastructure.io/public
{"jsonrpc":"2.0","result":"0x539","id":1}
```

aidan-starke:
I was a bit ahead of schedult - it has been updated now

```
curl -X POST -H  "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_chainId","params":[],"id":1}' https://cennznet.unfrastructure.io/public
{"jsonrpc":"2.0","result":"0x5359","id":1}
```

'0x5359' === 21337

ligi:
still - you cannot just do that as it opens the door for replay attacks. Why did you change the chainID in the first place?

aidan-starke:
chainId 1337 is registered as one of metamask's default networks so it cannot be added to metamask via `wallet_addEthereumChain`.

ligi:
> chainId 1337 is registered as one of metamask's default networks so it cannot be added to metamask via

interesting - do you know which chain it is there?

aidan-starke:
pretty sure it's used by ganache for local development

PR #1083, opened 2022-04-15
Remove Autobahn

defido:
The chain RPC URL and other factors will likely change, suggest it being added later. Have removed for now.


ligi (changes requested):
Please only add a deprecated field - we do not remove chains to protect against replay attacks

defido:
Ok, people will find out sooner or later that the RPC has changed.

ligi:
why don't just change the RPC to the new one then?

defido:
I won't be the maintainer. Someone else will need to do that.

PR #1102, opened 2022-04-23
Add Voltium Smart Chain

AntimonyIQ:


ligi (changes requested):
Your explorer does not conform to EIP3091 when it comes to address
Also the icon cannot be added like this - please look how the other chains do it

AntimonyIQ:
> 

Okay, let me work on this

PR #1103, opened 2022-04-24
Add eip155-62621.json for MultiVAC mainnet

0x13e:
I would be grateful if the MultiVAC chain could be added. The current explorer is custom, but a possible blockscout instance and integration with sourcify.eth in the near future.

### DIFF
--- a/_data/chains/eip155-62621.json
+++ b/_data/chains/eip155-62621.json
@@ -1,0 +1,24 @@
+{
+  "name": "MultiVAC Mainnet",
+  "chain": "MultiVAC",
+  "icon": "multivac",
+  "rpc": [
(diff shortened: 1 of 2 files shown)


ligi (approved):

Binary178:
How do I add a new network to my metamast account

Binary178 (commented):
I will be greatfull if polygon could be added

Stanley114:
The process is hard 
Send me your WhatsApp number so we can chat

Binary178:
On Mon, Apr 25, 2022 at 5:02 AM, Stanley114

07021331540

Stanley114:
Your number is not on WhatsApp

Binary178:
+2349021331540

PR #1107, opened 2022-04-26
Create FUNDING.yml

Elmzbitcx:


Elmzbitcx:
- name: Setup Java JDK
  uses: actions/setup-java@v3.2.0

ligi:
Not sure this is useful without any field actually filled out - when we have such an account applicable here we might do it - but for now I am just closing this

PR #1108, opened 2022-04-26
add explorers for dfk chains

MrFaun:
The following PRs add alternate explorers for the DFK Chain


ligi (changes requested):
I think we should remove the `"icon": "ethereum",` here

PR #1117, opened 2022-04-28
Create eip155-985.json

memolabsio:
add abbas  mainNet info


ligi (changes requested):
the explorer returns 502

memolabsio:
> 
the explorer  ok

PR #1125, opened 2022-05-04
Update eip155-777.json

genexsource:


ligi (changes requested):
This looks like a different chain - so this change is only possible if you prove your chain is older

PR #1142, opened 2022-05-09
Add SolarChainTestnet

likai1130:


ligi (changes requested):
I cannot access https://sctsan.starnet.ltd - also explorers need a standard field

likai1130:
Thank you for telling me why I failed. I want to ask, this is our internal test has not been opened, can it be merged?

ligi:
I do not think we should merge internal tesnets. What is the benefit having an internal testnet in a public repo?

likai1130:
We try to meet the standards first. At present, it is in the testing stage and will be open in the future.

likai1130:
It is mentioned in the document introducing the project that the internal chain can also be applied for. We first submit the JSON file to prevent chainId conflict, and then we will open the address after the test is completed.

ligi:
what document are you referring to?

likai1130:
I just found the document. It's this link https://geth.ethereum.org/docs/interface/private-network

ligi:
where does this document state a private chain can be registered here ?-)

likai1130:
In the title choosing a network ID
 "The main network has ID 1. If you supply your own custom network ID which is different than the main network, your nodes will not connect to other nodes and form a private network. If you’re planning to connect to your private chain on the Internet, it’s best to choose a network ID that isn’t already used. You can find a community-run registry of Ethereum networks at https://chainid.network/."
Can this sentence explain that private chains can be registered here?

ligi:
I do not think it states that private chains can be registered here - but yea one could interpret it that way.
Maybe we can create a field "private" that we can leverage here. But I really do not want that chains like that chains like this clutter sites like https://chainlist.org

ligi:
This needs a larger discussion - opened this issue: https://github.com/ethereum-lists/chains/issues/1144

likai1130:
OK, thank you very much for your patience. Then we open the chain to the outside world according to the current process, and then submit pr.

likai1130:
The chain has been opened. Let's see if it can be merged

likai1130 (commented):
add chainId 65535

PR #1170, opened 2022-05-20
Add a new network "excoincial"

SoneKatsu:


ligi (changes requested):
This ChangeSet does not look good

PR #1175, opened 2022-05-24
Add NxtMeta Blockchain Mainnet and Testnet

jsong1230:
Please add NxtMeta Blockchain Mainnet and Testnet.
Testnet will be available by July 1, 2022.


ligi (changes requested):
Please add a field `status` with the value `incubating`

Dong1999heo (approved):

PR #1178, opened 2022-05-25
 Added PolysmartChain network

371814861:
The polysmart chain is an EVM-compatible public blockchain.


ligi (changes requested):
The IPFS hash is ad directory - but must be a file

ligi (approved):
The CID must be a file not a directory

ligi (changes requested):
The explorer must have a standards field

PR #1179, opened 2022-05-25
suppress vscode plugin generated bin folder and files for kotlin proj…

masteraychen:
Within vscode IDE, kotlin and gradle etc plugin will generated .gradle and bin folders for kotlin projects, which are not needed for source code versioning. Hence, excluding them by adding the info inside .gitignore will be helpful for other folkers who are in the same situation.

### DIFF
--- a/_data/chains/eip155-160.json
+++ b/_data/chains/eip155-160.json
@@ -1,0 +1,23 @@
+{
+  "name": "Armonia Eva Chain Mainnet",
+  "chain": "Eva",
+  "rpc": [
+    "https://evachain.io/rpc/",
+    "wss://evachain.io/ws/"
+  ],
+  "faucets": [""],  
+  "nativeCurrency": {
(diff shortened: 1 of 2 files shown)


_data/chains/eip155-160.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "Armonia Eva Chain Mainnet",
+  "chain": "Eva",
+  "rpc": [
+    "https://evachain.io/rpc/",
+    "wss://evachain.io/ws/"
+  ],
+  "faucets": [""],
```
ligi, line 8:
```suggestion
  "faucets": [],  
```

ligi (changes requested):
I cannot resolve https://evascan.io

_data/chains/eip155-160.json
c4ei: oh source
"faucets": [], change ?

PR #1282, opened 2022-06-20
Updated PHI Symbol

Phinetwork:
Updated PHI Symbol ϕ to PHI because of error on web browsers to automatically add chain if symbol is less than 2 characters


Phinetwork:
Hello can we merge this @ligi

PR #1314, opened 2022-07-01
apply for  eth chainId

dxc520:


Truevato:
apply for eth chainId #1314

Truevato:
apply for eth chainId #1314

PR #1319, opened 2022-07-04
add aitd mainnet and update aitd testnet

moumc:
add aitd mainnet chainid 1319,  update aitd testnet chainid from 239 to 1320.


moumc:
Resubmit

PR #1322, opened 2022-07-04
Add Ambrosus Network

kiwvic:


ligi (changes requested):
missing field [faucets] - it can be an empty array - but it needs to be there

Randall10 (approved):

_data/chains/eip155-30746.json
```diff
@@ -0,0 +1,25 @@
...
+        "https://network.ambrosus-dev.io"
+    ],
+    "faucets": [],
+    "nativeCurrency": {
+        "name": "Amber",
+        "symbol": "AMB",
+        "decimals": 18
+    },
+    "infoURL": "https://ambrosus.io",
+    "shortName": "amb",
```
ligi, line 15:
```suggestion
    "shortName": "ambdev",
```

PR #1330, opened 2022-07-06
add OM Chain mainnet

cadcoin27:
add OM Chain mainnet

### DIFF
--- a/_data/chains/eip155-1246.json
+++ b/_data/chains/eip155-1246.json
@@ -1,0 +1,23 @@
+{
+	"name": "OM Chain Mainnet",
+	"chain": "omchain",
+	"network": "mainnet",
+	"rpc": [
+	"https://rpc-cnx.omplatform.com/"
+	],
+	"faucets": [],
+	"nativeCurrency": {
+		"name": "OMCOIN",
+		"symbol": "OM",
+		"decimals": 18
+	},
+	"infoURL": "https://omplatform.com/",
+	"shortName": "om",
+	"chainId": 1246,
+	"networkId": 1246,
+	"explorers": [{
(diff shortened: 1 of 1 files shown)


Keith1650 (approved):

nithronium:
@ligi the current version causes confusion for our user base as with the same name another network exists: omchain Mainnet - network ID: 21816 - symbol OMC

Is there a way to resolve the issue? Even though the symbol is OMC on ours, the networks' names are a direct match.

ligi:
I asked on twitter:
https://twitter.com/EthereumLists/status/1551588538822565891
first time this comes up for names - not yet sure how to deal with it. Please also chime in  @cadcoin27

nithronium:
I believe the reason this list exists is to remove the doubt from the end-user, and have a curated list. Project sharing same name - similar short names will cause confusion for the end user. For that reason I believe it is logical that we apply first come - first served principle here, and reserve the names for the networks, in our case name omchain (symbol OMC - ID: 21816) should be reserved. Including any short name that will take the form omchain. Because while our name is omchain, another project's short name being omchain is a problematic approach too. Otherwise no one stops anyone from creating another network named **Boba Base** and use the short name **bobanetwork** which would be confusing for the actual Boba Network users (https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-1297.json) as their network name is **Boba Network Bobabase** and shortname is **bobabase**

ligi:
the names where actually checked for uniqueness - but they where unique - one was called "omChain Mainnet" and the other one  "OM Chain Mainnet" 
addressing the problem in #1412 by normalizing the names (removing spaces and comparing case insensitive)

PR #1331, opened 2022-07-06
Update Optimism Mainnet info

K-Ho:
Updates the explorer link to the Optimism hosted explorer link, which currently proxies requests through to https://optimistic.etherscan.io/, but can proxy requests through to other explorers if Etherscan ever has issues or goes down.


ligi (changes requested):
Please add but do not replace
also why standard `none`  it's etherscan so it should support the EIP

PR #1332, opened 2022-07-06
Update Optimism Kovan Explorer info

K-Ho:
Updates the Kovan explorer link to the Optimism hosted explorer link, which currently proxies requests through to https://kovan-optimistic.etherscan.io/, but can proxy requests through to other explorers if Etherscan ever has issues or goes down.


ligi (changes requested):
Please add it as an alternative - but do not replace the old one

PR #1350, opened 2022-07-12
Add epee chain

dangminhnhut1993:


ligi (changes requested):
shortName MUST be unique

dangminhnhut1993:
@ligi Thanks for your checking, I updated shortName. Could you please help to check it again?

PR #1352, opened 2022-07-12
Add ETND Mainnet

ETND:


ligi (changes requested):
The Icon etnd does not exist - was used in eip155-131419.json

PR #1359, opened 2022-07-15
add Q networks

yurii-aristarkhov:


ligi:
Why is the ShortName of the TestNet Fisher?

yurii-aristarkhov:
> Why is the ShortName of the TestNet Fisher?

We ususally name testnets after famous scientists. Should it be like "q-testnet"?

ligi:
I think that would be better - more consistent - otherwise "Fisher" should also be in the name - not only shortName

ligi (approved):

PR #1369, opened 2022-07-17
Created eip-50000.json

AyzeGamingFR:
I actually created my own ethereum public chain, the goal of it is to host some projects like some NFTs projects, Tokens projects, Chain Bridges, etc.

The ethereums of the blockchain are entirely free and their price is of 0$, the goal is to get some fees of 0.

### DIFF
--- a/_data/chains/eip-50000.json
+++ b/_data/chains/eip-50000.json
@@ -1,0 +1,24 @@
+{
+    
+    "name": "AyzeLYC Public Chain",
+    "chain": "",
+    "rpc": [],
+    "faucets": [
+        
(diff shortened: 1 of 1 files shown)


ligi:
> and their price is of 0
how do you prevent spam / DOS of your chain?

AyzeGamingFR:
> > and their price is of 0
> > how do you prevent spam / DOS of your chain?
actually, the blockchain is in tests, im actually building a small team to check how much transactions can be sent each second to see how far does the nodes can hold the load, but it clearly won't be worse than the ethereum blockchain and there will probably never be as much transactions per second as inside the bsc or polygon blockchains

PR #1389, opened 2022-07-25
Update the Outdated Details for Cronos Mainnet

aw126:
Update to the latest Block explorer URL for Cronos Mainnet. 
Official Cronos RPC URLs page: https://docs.cronos.org/for-users/metamask


ligi (changes requested):
Seems to be 2 different explorers - so maybe just add it as an alternative? Or why does it need to be replaced?

aw126:
> Seems to be 2 different explorers - so maybe just add it as an alternative? Or why does it need to be replaced?

There are two explorers for Cronos chain: in-house blockscout explorer and Cronoscan. Blockscout has its own advantages and will not be replaced by Cronoscan based on the current plan. 
`https://cronoscan.com/` is applied as the main explorer in the latest RPC detail.

ligi:
yes - then please add it as a second explorer and do not replace the old one. It is an array because of this reason.

aw126:
> `https://cronoscan.com/` is more commonly used among Cronos users and applied as the main explorer in the latest RPC detail.

Though both explorers co-exist, from the perspective of RPC, the new explorer `https://cronoscan.com/` should replace the old one. The new one has now applied as the explorer URL in the latest RPC detail, so we can just keep the new one and remove the old one. Sorry for the confusion.

aw126:
Hi @ligi, another quick follow-up on this change request along with https://github.com/ethereum-lists/chains/pull/1391. May we know if there are any further questions/concerns blocking merging these two changes? Thanks.

ken-cronos:
@ligi 
hi -- friendly follow-up to update the Cronos chain details in chainlist according to the official endpoints. All the details below are the official ones

Mainnet
Name: Cronos
New RPC URL: https://evm.cronos.org
Chain ID: 25
Symbol:CRO
Block explorer URL:https://cronoscan.com/

Testnet
Name: Cronos testnet
PC URL: https://evm-t3.cronos.org
Chain ID:338
Symbol:tCRO
Block explorer URL: https://testnet.cronoscan.com/

_data/chains/eip155-25.json
```diff
@@ -12,11 +12,9 @@
...
-  "explorers": [
-    {
-      "name": "Cronos Explorer",
-      "url": "https://cronos.org/explorer",
-      "standard": "none"
-    }
-  ]
+  "explorers": [{
+    "name": "Cronos Explorer",
+    "url": "https://cronoscan.com/",
```
ligi, line 17:
```suggestion
    "url": "https://cronoscan.com",
```

ligi (changes requested):
Please run the prettier

aw126:
> Please run the prettier

Hi @ligi, we've run the prettier checks and it should be good now. Could you please review and merge?

ligi (approved):

PR #1401, opened 2022-07-30
Avalanche short name incorrect

matthewlilley:
Avalanche short name should not be "Avalanche", it shoud be "avax".


matthewlilley:
Gnosis overrides this as can be seen here:

https://gnosis-safe.io/app/avax:0x09842Ce338647906B686aBB3B648A6457fbB25DA

ligi (approved):

PR #1438, opened 2022-08-10
remove ask token eip155-222

gsalvador-permission:
we would like to remove our ASK token on your list


ligi (changes requested):
you can deprecate it - but we do not allow deletions as this opens the door to replay attacks

gsalvador-permission:
where can I deprecate it?

PR #1449, opened 2022-08-17
Add Dogechain and Dogechain Testnet

beeb:
Add Dogechain and Dogechain Testnet


Paty1283:
> Add Dogechain and Dogechain Testnet

Z👍🤩

PR #1453, opened 2022-08-18
Gnosis chains udpate

alebanzas:
- [x] Adding Chiado Testnet
- [x] Gnosis faucet, explorer, and docs URL
- [x] Naming


ligi (changes requested):
short name must be unique - but found `GNO` more than once

ligi (changes requested):
short name must be unique - but found `GNO` more than once

ligi (approved):

PR #1454, opened 2022-08-18
depricate ASK permission token

gsalvador-permission:


ligi (approved):

gsalvador-permission:
@ligi We would like to remove our legacy blockchain listing from Chainlist, I was instructed that for me to do it, I should remove our list at this Repo, but it's still reflecting at Chainlist. any suggestion?

PR #1458, opened 2022-08-19
Added Canto Mainnet

ColinPlatt:
Please


ColinPlatt:
removed missing field [faucet]

ligi (changes requested):
filename extension must be json

PR #1460, opened 2022-08-20
update cndl

samisbakedham:


ligi (changes requested):
short name must be unique - but found `CNDL` more than once

samisbakedham:
What do you mean?

PR #1470, opened 2022-08-21
Try fix rebase workflow

ligi:
https://stackoverflow.com/questions/66643917/refusing-to-allow-a-github-app-to-create-or-update-workflow


jonesk7734 (approved):
Ok

PR #1476, opened 2022-08-22
add etf

FranckStone:


ligi (commented):
not a number at networkId

ligi (changes requested):
ShouldHaveNoMissingFields: missing field(s) [shortName]

ligi (approved):

PR #1499, opened 2022-08-27
Update eip155-8080.json with Shardeum Network

ashutoshpw:
- Removed the older Network as it is down from a long time
- Added Shardeum network which has the same chain ID

### DIFF
--- a/_data/chains/eip155-8080.json
+++ b/_data/chains/eip155-8080.json
@@ -1,17 +1,29 @@
 {
+<<<<<<< HEAD
   "name": "GeneChain Adenine Testnet",
   "chain": "GeneChain",
   "rpc": ["https://rpc-testnet.genechain.io"],
   "faucets": ["https://faucet.genechain.io"],
+=======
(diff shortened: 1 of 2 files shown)


ligi:
I need to add a flag that indicates this is a reused chainID as this has the risk of replay attacks

ashutoshpw:
Should I add `isChainIdResued:true` or do you have something else in mind?

ligi:
You cannot do it yet - I need to add it to the CI - otherwise it will fail.
Also currently I am leaning more to something generic like "redFlags":["reusedChainId"]

ashutoshpw:
Got it, thanks.

ligi:
redFlags is now supported

manny363:
If your trying to help me out Bless you and if your not it's really not the
best time for this. Like I told you I have had a long stressful week and
half .

ashutoshpw:
Ready to be merged now.

jonesk7734 (approved):
Ok

Issue #610, opened 2021-11-29
validate shortName


ligi:
should not contain a space and have some maximum length

PR #1509, opened 2022-08-28
Check and validate JSON schema

ashutoshpw:
This will check for a valid JSON schema on every push and PR request to the repository. 

Schema Definition is located in 
`tools/schema/chainSchema.json`


ashutoshpw:
@ligi , I have added following params as required in the JSON field `"name","chain","rpc","nativeCurrency","chainId","networkId"`

Do you think any other field in the JSON is required?

ligi:
you can see all I see as mandatory here: https://github.com/ethereum-lists/chains/blob/9e35cab9b0d284acd0c5e71a955e57e4399d117a/processor/src/main/kotlin/org/ethereum/lists/chains/Env.kt#L11

ashutoshpw:
Adding a Regex definition in `chainSchema.json` would also solve this issue. https://github.com/ethereum-lists/chains/issues/610 . `shortName` in 19 files needs to be fixed if we use the regex `^[A-Za-z0-9-_]+$` to allow alpha-numeric characters without space.

Keith1650 (approved):

alecodesign:
> This will check for a valid JSON schema on every push and PR request to the repository. 
> 
> Schema Definition is located in 
> `tools/schema/chainSchema.json`
> 
>

jonesk7734 (approved):
👍

PR #1522, opened 2022-08-30
Add Opside Testnet

icydark:
Add opside Testnet, chain ID 23118


ligi (changes requested):
The icon opside-test does not exist

PR #1524, opened 2022-08-30
Generate Icon list JSON in build

ashutoshpw:
As of now, we can list the chains using the data from the build but there was no way to list all the icons. 

This PR generates `chain_icons.json` and `chain_icons_mini.json` at the build time so we can list down all the icons that were defined in `_data/chains/eip155*.json`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,9 @@ _site
 .env
 node_modules
 .idea
+output
+.gradle
+model/bin
+model/build
+processor/bin
+processor/build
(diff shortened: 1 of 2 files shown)


ligi:
thanks - but I (currently) fail to see a use-case for such a file

ashutoshpw:
Let's say there is tool where we list down Chains from `chains.json` and filter out few chains on basis of some rules. Suppose we have 30 chains in the filtered list. Now if want to show them over a dropdown with icons, we have to make 30x2=60 different calls to fetch the icon's json file followed by the call to IPFS network to fetch the exact icon. 

If we have this chain_icons.json then we can get the icons in 31 calls 
- 1 call to get the `chain_icons.json`
- 30 calls to load the icons from IPFS network.

ligi (approved):

PR #1535, opened 2022-09-02
Request to add new chain

heyoka-448:
Please add Credit Smartchain Mainnet and Testnet


ligi (changes requested):
shortName must be unique

PR #1544, opened 2022-09-05
Create eip155-10002.json

NeoV55:
ETHW-iceberg-testnet


ligi (changes requested):
the networkId is missing - also please run the prettier

PR #1548, opened 2022-09-06
Revert "Another approach to fix auto rebase"

Apdlrcjafg19:
Reverts ethereum-lists/chains#1513


ligi:
I am all for fixing the rebase - but this will not work - tried it

PR #1557, opened 2022-09-07
Replace rpc node path

steveliupai:
Replace lambda testnet rpc node


steveliupai:
@ligi Please merge, thank you so much!

PR #1565, opened 2022-09-08
Add Planet Memes Chain testnet

headria:


ligi (changes requested):
where did you pin the IPFS CID? I could not resolve it via `ipfs get`

PR #1578, opened 2022-09-11
fix: shardeum 1.3 is down and now replaced with shardeum 1.4

kgmyatthu:
Shardeum 1.3 is down and shardeum 1.4 is now replaced on it with the same rpc, explorer urls and chainID.

### DIFF
--- a/_data/chains/eip155-8080.json
+++ b/_data/chains/eip155-8080.json
@@ -1,5 +1,5 @@
 {
-  "name": "Shardeum Liberty 1.4",
+  "name": "Shardeum Liberty",
   "chain": "Shardeum",
   "rpc": ["https://liberty10.shardeum.org/"],
   "faucets": ["https://faucet.liberty10.shardeum.org"],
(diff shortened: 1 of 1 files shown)


_data/chains/eip155-8080.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "Shardeum Liberty 1.3",
+  "name": "Shardeum Liberty 1.4",
```
ligi, line 2:
```suggestion
  "name": "Shardeum Liberty",
```

ligi (changes requested):
Let's just remove the version number from the name - then we do not need to update the chain name every time the version bumps

_data/chains/eip155-8080.json
kgmyatthu: Hi we have multiple test net running at a time with the milestones and different set of feature and version number are only the could differentiate.

kgmyatthu: We currently have 1.4 and 2.0 running live right now under liberty namespace.

ligi (approved):

PR #1602, opened 2022-09-16
Add Ethereum PoW

ChrisiPK:
Correct filename is unfortunately already taken by Smart Bitcoin Cash. Please suggest how this should be named instead.

### DIFF
--- a/_data/chains/eip-10001.json
+++ b/_data/chains/eip-10001.json
@@ -1,0 +1,21 @@
+{
+  "name": "Ethereum Proof of Work Mainnet",
+  "chain": "ETH",
+  "rpc": [
+    "https://mainnet.ethereumpow.org"
+  ],
+  "faucets": [],
+  "nativeCurrency": {
+    "name": "Ether PoW",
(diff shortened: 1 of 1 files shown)


_data/chains/eip-10001.json
```diff
@@ -0,0 +1,21 @@
...
+    "https://mainnet.ethereumpow.org"
+  ],
+  "faucets": [],
+  "nativeCurrency": {
+    "name": "Ether PoW",
+    "symbol": "ETHW",
+    "decimals": 18
+  },
+  "infoURL": "https://ethereumpow.org/",
+  "chainId": 10001,
```
pfedan, line 14: #1673

ligi:
Ethereum PoW will never be listed here with chainID 1001 - chainID duplications expose users to replay attacks - this is a no go. And in this case they where even warned berfore.

tallero:
I see Ethereum PoW documentation still suggests using that ChainID.
Has anything changed since that time?

KingOfNorway (commented):
Ye

PR #1621, opened 2022-09-19
Update Dexalot Subnet

connorch:
I'm with the Ava Labs team.  This PR is to update the Dexalot subnet to match our subnet branding guidelines, fix the faucet URL, and add an icon.


connorch:
@ligi Can you please take a look at this?  Thanks!

PR #1625, opened 2022-09-20
Update portal fantasy subnet

connorch:
Hello, I'm from the Ava Labs team.  This is a PR to update the portal fantasy subnet to align with our branding, as well as provider a higher res icon for the chain.


ligi (changes requested):

PR #1629, opened 2022-09-21
Add Bitkubchain Mainnet

ppanya:
Added Bitkub Chain, Replace Next Smart Chain (Old chain same chain id: 96)

Detail Genesis Block Next Smart chain: [Block 0](https://explorer.nextsmartchain.com/blocks/0)

- Mainnet
  - Network ID: 96
  - Detail Genesis Block: [Block 0](https://www.bkcscan.com/blocks/0/transactions)


ppanya:
@ligi sorry, I want to know what the average time for pull request review or any progress?

macvts:
@ligi can you provide some assistance toward this request?

ligi (changes requested):

dcnl1980:
Chain Id 96 is already in use for next smart chain, it will conflict. You can use another chain id for the mainnet.

dcnl1980:
Before next smart chain we run on another blockchain which is even older. I will send proof of this genesis block.

ligi:
> Before next smart chain we run on another blockchain which is even older. I will send proof of this genesis block.

this would help - then this is likely not to be merged here - except for the Testnet - @ppanya - please split in this case

macvts:
would like @dcnl1980 to share the proof of genesis block here so thats can be right for @ligi to consider in this case. 

Some of my opinion: if there is another blockchain with older genesis block using chain id 96, isn't it need to change into that name of the blockchain? 

tbh if chain id 96 will be submitted as a NEXT Smart Chain, Bitkub Chain has the right to claim this due to the older genesis block.

macvts:
on top of that, next smart chain currently have around 25.9K blocks with 5,342 transactions while Bitkub Chain exceeded 10M blocks with total transaction of 339M as of now. 

Isn't it suitable for NEXT Smart Chain to be the one who change the chain ID? 

for comparison:
www.bkcscan.com
explorer.nextsmartchain.com

macvts:
@ligi gently follow-up for your reconsideration toward this case

ligi:
@dcnl1980 plese send the proof - otherwise I need to merge this one (but it for sure needs `"redFlags": ["reusedChainId"]` then)
I give you 7 days to respond - otherwise it will be merged (with redFlags)

dcnl1980:
@ligi Proof will be send later today.

dcnl1980:
> @ligi Proof will be send later today.

In the attachment you find the proof of the old blockchain, which we run before NEXT Smart Chain and which have the same Chain ID 96.

The genesis block was mined with timestamp 1555502400, which is Wed Apr 17 2019 12:00:00 GMT+0000. 

NEXT itself has been active in the market since October 2017.

> on top of that, next smart chain currently have around 25.9K blocks with 5,342 transactions while Bitkub Chain exceeded 10M blocks with total transaction of 339M as of now.

Regarding the above I want to object that the amount of blocks or transactions per blockchain can be different, the old NEXT Chain had over 1,8M blocks, but the new NEXT Smart Chain (more efficient) only creates blocks when transactions are actually performed and not every few seconds like BitKub does.

**Thinking in solutions. I have no problem with both names being listed under Chain ID 96 (if possible). It would solve both problems. Changing a chain ID for a running public and operative blockchain is almost impossible without starting a new one.** 

screenshot-16012023105512

macvts:
I'm not quite sure if the proof you provide is valid or not, is it right to use this proof of old blockchain (inactive or currently active?) as an authority to chain id 96 is not suitable at all as I mentioned in my thought above.  

Also speaking of the discussion on the genesis block on the current NEXT Smart Chain, we didn't ask for NSC to change your chain ID as you said before but we only need the right the be presenting as the only blockchain on chain ID 96 user's search 

@ligi please considerately suggest in this case whether it's right to use the old blockchain genesis block (which is unknown for the current status) to claim the authority to represent the new chain which is younger than Bitkub Chain.

ligi:
@ppanya please add `"redFlags": ["reusedChainId"]`
@dcnl1980 unfortunately this is not really a proof and could be crafted - so I think I have to go with merging as for our [collision management policy](https://github.com/ethereum-lists/chains#collision-management). Why didn't you choose a new chainID back then? This would have been much better (also wrt replay attacks)

ppanya:
@ligi `"redFlags": ["reusedChainId"]` has been added

ligi:
dam this is messy - our current policy is oldest genesis wins. But I cannot just take a screenshot - everyone could just fake this. But if the chain is still active this changes things. 
Also tempted to change it to "first one registering wins" - which makes it much clearer and less prone to ugly discussions like this.

macvts:
To clarify, I myself and @ppanya didn't plan to make to discussion go this way since we're considering the current policy as the most important factor, there can be some questions for @dcnl1980 regarding our actions but we're really open, and didn't mean to disrespect in any case.

I did understand how frustrating this could be for @ligi and we're very sorry if this case cause a mess for you and will respect your attentive decision. 

on the other hand, if NEXT Smart Chain can provide their proof of the old chain's genesis block, will NSC be able to connect that one to the chainlist? since you're claiming the right of this using the old chain and we'll not doubt anything from that.

> Chain Id 96 is already in use for next smart chain, it will conflict. You can use another chain id for the mainnet.

> Changing a chain ID for a running public and operative blockchain is almost impossible without starting a new one.

TL;DR these 2 opinions above, seem like you're finding the space for an objection at the stage while you're rushing us to do the almost impossible task 2 days ago.

ligi:
https://twitter.com/EthereumLists/status/1615056830278864903

I need to think about this situation a bit more and get more input on the general situation.

macvts (changes requested):
@ppanya can you update the details of the explorer name from bkcscan into Bitkub Chain Explorer? and also some recommendations to use only Bitkub Chain as a name of the chain without adding terms like Mainnet to it just like Ethereum, Avalanche, Optimism, Arbitrum One

on the other hand, you can clarify the Testnet as a Bitkub Chain Testnet, and Bitkub Chain Testnet Explorer. #2145

macvts (approved):

macvts:
> @dcnl1980 plese send the proof - otherwise I need to merge this one (but it for sure needs `"redFlags": ["reusedChainId"]` then) I give you 7 days to respond - otherwise it will be merged (with redFlags)

I would like to follow up on this case after a week passed by since the response of proof is not actually legit, and there is no further proof of it as well. I wanted this case to be considered as neutral as possible upon the current policy. @ligi

ligi:
I will very likely merge this soon - seems the current policy and arguments are on your side - but first I really need to change the policy. Never want to be in this situation again.

dcnl1980:
@ligi we send enough proof about our earlier existence and usage of chain id 96. Since this chain is still up and running we reserve the right to put the block explorer online and therefore claim the chain id 96 as the oldest block chain.

ligi:
I do not think it is enough proof yet - this screenshot is trivial to create. Doing a chain with blockexplorer is at least a bit harder - but yea I know it can also easily be faked (at least for non PoW chains) - but so could the other chain be.

But to not be in the situation again I now chainging the policy:

https://github.com/ethereum-lists/chains/pull/1629#issuecomment-1400326642

ligi:
After Explorer and RPC stand the trial of some tests I will close this PR

dcnl1980:
Got it, but this was the previous blockchain after it was merged to another one. So we need time to put the old explorer online (since we take it offline). Anyway you can push Bitkub further. In a later stage I will come back to overrule them by this older blockchain. In that case I want to make sure that it will not make any issue. This change will be just temporary.

dcnl1980:
@macvts  make it clear. We have all the right to bring the old blockchain up. You had to register it a long time ago. You awake after 2 years to claim chain id 96, while we are in the market since 2017, that is a fact. You also called it Bitkub NEXT, so it can be also a trade mark issue as well. Our blockchain will overrule the one which you have if we put the explorer online. You have no right to take chain 96 at all. You are trying to suggest that your blockchain have more transactions, but that is also not a valid proof. Reject this Bitkub thing, on the first serve, first come. We are already in the market since 2017. It is not our mistake that Bitkub comes with a claim after several years.

macvts:
> while we are in the market since 2017, that is a fact.

on the other side of your fact, NSC is the one who starts to split the new blockchain to restart the whole structure (younger genesis block) which you should be aware of the collision policy that the oldest genesis block will have the authority toward the chain ID 96 and now you resisting while trying to use your no-use blockchain (or maybe unknown) to claim the authority instead. FYI @ligi 

currently to the case, we're considering while the policy with the oldest to win and some info that might be more useful before any slander to anyone with a trademark issue @dcnl1980 
Bitkub Chain = blockchain
Bitkub NEXT = Digital Wallet Service Provider
Bitkub Exchange = Crypto Exchange
Bitkub Venture = Venture Capital
Bitkub Academy = Educational Hub for Crypto Trader

dcnl1980:
@macvts we have the rights to upgrade our blockchain, I don’t see any issues in that. Even with our older blockchain (which is still operational and in use) we are much older than BitKub.

ligi:
I am very much leaning this way after thinking about it for a while:

image

https://twitter.com/danfinlay/status/1617597043949830144

macvts:
will totally respect your decision, I also hope somehow that this could be work for Bitkub Chain since we never go against any rule or issue from the very beginning.

macvts:
> I will very likely merge this soon - seems the current policy and arguments are on your side - but first I really need to change the policy. Never want to be in this situation again.

> After Explorer and RPC stand the trial of some tests I will close this PR

Gentle follow-up for the direction of the case, I genuinely provide a total of 3 weeks for NSC to provide proof to support their authority.  probably want this to be fairly considered from the beginning of the discussion with the past policy that just recently changed due to the confliction of this case. @ligi

dcnl1980:
Not sure what the issue is. We are still the oldest chain for over a few years. It is not our issue that BitKub did not register their ChainID on time. We check your website and it was available.

Secondary we are still the old blockchain which can overrule BitKub by being the oldest. This chain is still online and if needed we will release the explorer to prove it, but it will take some time, since the old chain is depreciated. 

So I would like to make sure and agree with you that we can claim this id, if we are the oldest chain, not that you are changing the rules in between. At all we will not change our id at all and keep running on chain id 96.

ligi:
@dcnl1980 so far you did not provide proof - you where asked to provide provide explorer and RPC multiple times - and you did not.
I think the best is to go with Dan Finlays suggestion and just remove the chain
PR #2251 opened - will likely be merged soon
You can then add back your *old* chain if you want.
I think this is the safest option for users while still following the policy from this time.

dcnl1980:
> @dcnl1980 so far you did not provide proof - you where asked to provide provide explorer and RPC multiple times - and you did not.
> I think the best is to go with Dan Finlays suggestion and just remove the chain
> PR #2251 opened - will likely be merged soon
> You can then add back your _old_ chain if you want.
> I think this is the safest option for users while still following the policy from this time.

as already stated, the current blockchain is an update of a much older blockchain (older than BitKub) which is set to chain id 96, however this old blockchain is still operational. 

As mentioned by you I will reserve the right to claim chain id 96 in a later stage, when the RPC and explorer are online.

ligi:
Wondering how the old chain is still operational when you cannot provide me with an RPC. If the chain is still operational it should not take long to just expose an RPC of a node.
Also wonder what you where thinking reusing a chainID for something that does not seem to be just a TestNet - how do you protect your users from replay attack?

dcnl1980:
> Wondering how the old chain is still operational when you cannot provide me with an RPC. If the chain is still operational it should not take long to just expose an RPC of a node.
> Also wonder what you where thinking reusing a chainID for something that does not seem to be just a TestNet - how do you protect your users from replay attack?

We disable the public RPC and explorer, but there are still miners active on this network which are directly connected on their own RPC node interface. 

A reply attack is not possible, each blockchain have another consensus protocol. We choose to continue with the chain id since we defined it before in the genesis file and it is most logical since our customers are connected to it with their clients.

Anyway we hoped we did not had to open the public rpc and explorer again, but we are forced to do so. I asked the devs to put it back, after that is done, I will come back to claim chain id 96.

ligi:
> A reply attack is not possible, each blockchain have another consensus protocol.

how exactly does the consensus protocol change prevent replay attacks?

ligi:
I just merged #2251 - @dcnl1980 you might re-add your old chain answering above question and provide RPC/explorer for the old one.

macvts:
will this PR be considered in any scenario in the future? or @ligi will only wait for their answer to re-add or no one get this chain ID?

ligi:
yea - will likely be merged as @dcnl1980 is not answering anymore - so seems the case is leaning on your side now. But u need to rebase or merge to master

ligi (approved):

PR #1636, opened 2022-09-23
update bttc testnet

xq-lu:
update bttc testnet chain info.includes:chainid,rpc url, faucets


ligi (changes requested):
If the chainID changes it is a new chain. You can deprecate the old one - but not just change values

xq-lu:
close it

PR #1648, opened 2022-09-26
Boba Netowrk shortName to adhere to EIP-3770

matthewlilley:
Boba Netowrk shortName to adhere to EIP-3770


matthewlilley:
@ligi unsure why prettier is failing, but another shortname lowercase

matthewlilley:
@ligi forced re-run of prettier. Seems good now!

PR #1649, opened 2022-09-26
Add Bobascan explorer

matthewlilley:
Add Bobascan explorer

### DIFF
--- a/_data/chains/eip155-288.json
+++ b/_data/chains/eip155-288.json
@@ -17,7 +17,7 @@
     {
       "name": "Bobascan",
       "url": "https://bobascan.com",
-      "standard": "EIP3091"
+      "standard": "none"
     },
     {
       "name": "Blockscout",
--- a/_data/chains/eip155-288.json
+++ b/_data/chains/eip155-288.json
@@ -15,6 +15,11 @@
 
   "explorers": [
     {
+      "name":"Bobascan",
+      "url":"https://bobascan.com",
+      "standard":"EIP3091"
+    },
+    {
       "name": "Blockscout",
       "url": "https://blockexplorer.boba.network",
       "standard": "none"
--- a/_data/chains/eip155-288.json
+++ b/_data/chains/eip155-288.json
@@ -15,9 +15,9 @@
 
   "explorers": [
     {
-      "name":"Bobascan",
-      "url":"https://bobascan.com",
-      "standard":"EIP3091"
+      "name": "Bobascan",
+      "url": "https://bobascan.com",
+      "standard": "EIP3091"
     },
     {
       "name": "Blockscout",


SsNiPeR1 (approved):
ok.

SsNiPeR1:
wait, I need to check the compliance with eip

matthewlilley:
> oh, this is made by etherscan, no questions then

Yeah, regular Etherscan.

ligi (changes requested):
The explorer seems to be not EIP3091 compatible - just tried block an it 404s - either fix it or change to `none`

matthewlilley:
> none

Ah, I assumed all Etherscan clones are EIP3091. I'll share this with the team.

matthewlilley:
@ligi set to none for now. I'll open a follow up PR if they add the compatibility routes.

matthewlilley:
@ligi quick follow up on this one:

I double checked with other L2 explorers like optimism which is listed as EIP-3091 also, but being an L2 the block route is not present whereas batch route is. So it would also 404 for block routes.

https://github.com/ethereum-lists/chains/blob/313b649b2c4ecbc5500284db49d74175559833bd/_data/chains/eip155-10.json#L20

Most dApps don't make use of the block route, so it doesn't affect our app. 

Maybe a new EIP is needed which covers L2s use of batch which seems to replace block in these cases.

SsNiPeR1:
> Most dApps don't make use of the block route, so it doesn't affect our app.

it does: when you specify the standard, you must follow it.
you can edit standard parameter to "none" if it does not comply, or modify this explorer, or use some other explorer like [mine](https://github.com/SsNiPeR1/explorer).

SsNiPeR1:
> > none
> 
> Ah, I assumed all Etherscan clones are EIP3091. I'll share this with the team.

yes, i thought the same

SsNiPeR1 (approved):
now everything is ok

ligi (approved):

SsNiPeR1:
@ligi approve my pr now pleeeeeease

matthewlilley:
> which



> > Most dApps don't make use of the block route, so it doesn't affect our app.
> 
> it does: when you specify the standard, you must follow it. you can edit standard parameter to "none" if it does not comply, or modify this explorer, or use some other explorer like [mine](https://github.com/SsNiPeR1/explorer). upd: my explorer is not so stable and it has some bugs. i am just too lazy to fix it.

Yeah. I agree, I think this is a problem with lack of a sister standard for L2 Etherscan-like explorers which all seem to use the batch route instead of block. As pointed out in previous message, Optimism Etherscan has a type of EIP-3091 but actually isn't compliant. If this were to change now, it would break our application tx links etc..

PR #1658, opened 2022-09-27
Add uptick testnet chain

YangChuantongStarry:
Add uptick testnet chain


ligi (changes requested):

also: Native currency symbol must have less than 7 chars

PR #1667, opened 2022-09-29
Add basic entry for BTC

FrederikBolding:
Add basic entry for BTC to verify how CI works for non-EIP155 CAIP-2 chains.


SsNiPeR1:
what the hell is it... 
image
do you see bitcoin there in name? ;)
<b>BITCOIN IS NOT SUPPORTED</b>

SsNiPeR1 (changes requested):
delete this cringe

PR #1671, opened 2022-09-30
Add Agora testnet chain infos

TrustHenry:
Please.


ligi (changes requested):
the chainID seems to be used - and seems the other chain is older - so this is a non merge - or can you prove your chain is older?

PR #1675, opened 2022-09-30
Revert back to previous Prettier implementation

FrederikBolding:
Reverts back to previous Prettier implementation using `npx prettier` directly.


SsNiPeR1 (changes requested):
you should document all changes, you have changed a chain file too

jonesk7734 (approved):
😃

PR #1680, opened 2022-10-01
Disallow leading zeroes in file names

### LINKED ISSUE #1554: Check that there is no leading zero
image

FrederikBolding:
Expands the schemacheck code to be more strict when parsing chain ids from file names. Furthermore, solved the conflict between Factory 127 testnet and Lycan chain, removing F127 as there was no trace of it existing online.

Fixes #1554


ligi (approved):
Thanks!

SQuIsHY7D3m (changes requested):
Redundancies and multiple fails

PR #1702, opened 2022-10-07
rename ETH to Ethereum

LufyCZ:


matthewlilley:
It seems use of "chain" varies depending on the network. 

One thing we notice is that "name" field is sometimes not human friendly, e.g. Harmony where it's "Harmony Mainnet Shard 0"

https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-1666600000.json#L2

However it has a human friendly "chain" property with value "Harmony".

Ethereum however has a "name": "Ethereum Mainnet" and "chain": "ETH".

Would be good to get some clarification on what is best for these.

PR #1696, opened 2022-10-06
Add wagmi compatible chain listings

ligi:
closes #1683

PR #1712, opened 2022-10-11
Revert "Add wagmi compatible chain listings (#1696)"

AZIMUT98155:
This reverts commit 31b93d98adbe6310242b290e1e8fa034e5ce5077.


ligi (changes requested):
you need to explain why

PR #1736, opened 2022-10-19
Updated Metis Testnet Rinkeby chain to Goerli

psinelnikov:


ligi (changes requested):
Please do not remove chains - you can set `"status": "deprecated"` - but they can never be removed (for replay attack reasons)

Saberhabeelshohdy (approved):

ligi (changes requested):

ligi (approved):

PR #1753, opened 2022-10-21
add bronos testnet

chedetinaveen:


ligi (changes requested):
Can you please run the prettier?

PR #1780, opened 2022-10-26
add memo chain and icon

wy-h:


ligi (changes requested):
please run the prettier

PR #1792, opened 2022-10-28
Chiado Testnet reset

alebanzas:
- Deprecate old Chiado chain, id 100100
- Added new Chiado chain, id 10200


ligi (changes requested):
Do not delete old chains - only mark them as deprecated - this is important to protect replay attacks.

alebanzas:
Thanks @ligi for your review. It's now good to go!

PR #1820, opened 2022-11-03
Add chain frenchain

shopglobal:
_Testnet_:
**chainID**: 69420
**networkID**: 69420

_Mainnet_: 
**chainID**: 420
**networkID**: 69420

Mainnet is offline, but launching in some days. Thanks,M


ligi (changes requested):
the filenames are wrong

shopglobal:
@ligi the filenames have been corrected. Our intention was to avoid collision with other (seemingly defunct) chains containing similar filenames.

ligi:
This is not possible - you have a collision here - and the other chain on 420 is older  - so yours will not be listed

shopglobal:
> This is not possible - you have a collision here - and the other chain on 420 is older - so yours will not be listed

Well I hear you and I see you but my request is that we consider merge anyway since the network ID of this chain is not the same to the other chain ID 420.

ligi:
no chance - having the chainID being the same allows replay attacks - will not merge

shopglobal:
Ok will get them changed. Thanks,M

PR #1501, opened 2022-08-27
Added Prettier and formatted existing JSON files

ashutoshpw:
To keep the formatting uniform for all the JSON files that we have, I have configured prettier for JSON file and formatted all the files with prettier. 

CI/CD Command to run before merging for new PRs
`npx prettier --check _data/`

PR #1826, opened 2022-11-05
Remove error claim of Ganache default port and network ID (GUI)

rickmark:
Someone created a network at the address used by tons of Eth developers using the Ganache GUI.  This should be reverted and the Ganache system should claim that chain.


rickmark:
Reverts an error chain addition (takeover of default Ganache IDs) from #1501

_data/icons/ganache.json
```diff
@@ -0,0 +1,9 @@
+[
+    {
+      "url": "ipfs://Qmc9N7V8CiLB4r7FEcG7GojqfiGGsRCZqcFWCahwMohbDW",
+      "width": 267,
+      "height": 300,
+      "format": "png"
+    }
+  ]
+
```
ligi, line 9:
```suggestion
```

ligi (changes requested):
Thanks for spotting this!

PR #1827, opened 2022-11-05
Correct mis-allocation of 1337 from Ganache CLI to defunct network

rickmark:
Another case of a network claiming a common development configuration raising all kinds of grief for ETH developers

Reverting 1337 to ganache CLI default


rickmark:
Closed as 1337 is better as Geth private

PR #1834, opened 2022-11-05
Automatic Format / Test `pre-commit` Hooks (Husky / Node based)

rickmark:
Since I saw that node was used for some of the format / schema I added and configured `husky` to ensure that devs get format / tests performed pre commit (reducing the number of PRs that get kicked back for silly reasons)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ model/bin
 model/build
 processor/bin
 processor/build
+tmp/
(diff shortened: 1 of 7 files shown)


rickmark:
The build hosts would require 'graphicsmagick' for the new image validator to work correctly, I think its worthwhile to add as a apt install

ligi:
Thanks! But the pre-commit hook will only affect people that have husky installed right? Need to do some experiments after the CI for it is green. Bit scared about IPFS in the CI flow though - guess this will make a lot of problems - should also fallback to the iconDownload folder.

PR #1841, opened 2022-11-07
Update eip155-24.json with Kardiachain Network

phongnhat19:
- Removed the older network (Dithereum) as it's mainnet is not live for a long time.
- Added Kardiachain network which has the same chain ID


phongnhat19:
@ligi Thanks for the tweet, I've already reached out to Dithereum's Telegram channel but no response yet.
What is your detail process of replacing a chain? I can't find information on that.

Lan-Phan:
Hello @ligi ,

Thanks so much for your reference document.
For your information, definitely we're the one who take that chain id because:
+ Dithereum still doesn't have main net (their mainnet uses chain id 24)
+ Our kardiachain mainnet launched 684 days ago (Tue, 29 Dec 2020 12:00:00 GMT), check here: https://explorer.kardiachain.io/block/1

Please help to review, thanks.

phongnhat19:
@ligi It's been several days and still no response from Dithereum team, can you help us take a look and merge our pull request?

phongnhat19:
Hi @ligi , is there anything else needed for the PR? If no, can you help us continue the process.
Thanks.

ligi (changes requested):
where did you pin the IPFS CID? I could not resolve it via `ipfs get`

also the icon json does not contain the size

phongnhat19:
@ligi just re-upload the logo to IPFS and update `width` and `height` to icon json. Can you help us recheck? Thanks.

PR #1852, opened 2022-11-09
update incorrect name and path in prev commit of new json file

cnukaus:
Previous commit had the json file in root folder with a wrong name. now moved to _data\chains\eip155-3666.json


ligi (approved):
approving and will merge - but you might want to bring back the icon

PR #1857, opened 2022-11-09
chainId 8080 Shardeum Liberty 1.6

MarcusWentz:


ligi (approved):
approving - but really hate that the version number is in there tbh - just makes me work for nothing - would really like that the version number is not in there

MarcusWentz:
I can change 1.6 to 1.X and 2.0 to 2.X to 
avoid new version number updates if we change to 1.7 to 2.1 [I will open a PR for this in this scenario to prevent more PRs].

PR #1858, opened 2022-11-10
changed the MOAC chain rpcs to https

mdranger:
changed the testnet 201 and mainnet 1099 rpcs to https


mdranger (commented):
changed the rpcs from http to https

ligi (approved):

PR #1871, opened 2022-11-14
Add Dehvo Chain ID 113

Phinetwork:
Add Dehvo Chain ID 113


Phinetwork:
Do not understand why run prettier failed when it has been added and passed before with same syntax

Phinetwork:
Opened New Pull Request With Fix

PR #1882, opened 2022-11-17
add qitmeer icon

blocklee:


ligi (changes requested):

blocklee:
> tl;dr: it is now required to put the binary icon into iconDownload
> 
> why: it was a PITA to always `ipfs get` in the review process - sometimes it takes _really long_ or even fails (especially with pinata) - this took significant time (which adds up with the amount of chains added) and slowed down the review process. Also sometimes CIDs become unavailable after - big mess. Hence it is now required to put them in iconDownload (still named via CID)

done, thanks.

ligi (approved):

PR #1883, opened 2022-11-17
create psc network

ssspsa:
In the PolySmartChain network, there is a normalized computer (called the virtual machine, or EVM) whose state is agreed upon by everyone in the PolySmartChain network. Each participant in the PolySmartChain network (each PolySmartChain node) keeps a copy of the state of this computer. In addition, any participant can broadcast a request to this computer to perform any computation. Whenever such a request is broadcasted to the network, the other participants on the network will verify, confirm, and perform ("execute") the computation. This command causes a change in the state of the EVM and is propagated throughout the network.
The computation request is called a transaction request; the record of all transactions and the current state of the EVM are stored in the blockchain, which in turn is stored and agreed upon by all nodes.
The encryption mechanism ensures that once a transaction is verified as valid and added to the blockchain, it cannot be tampered with afterwards. The same mechanism also ensures that all transactions are signed and executed with the appropriate "permissions" (no one can send digital assets from Alice's account except Alice herself).


ligi (changes requested):

ssspsa:
Resubmitted

milad237 (commented):
invitr

ssspsa:
Please review

ssspsa:
@ligi Hi, please review and merge, thank you so much

ligi (approved):

PR #1891, opened 2022-11-18
Mark PlatON Devnet as deprecated

benbaley:
PlatON devnet are no longer in use now, we mark it as deprecated.


ligi (changes requested):
you can mark it deprecated - but not remove it - to protect replay attacks

ligi (approved):

benbaley:
> you can mark it deprecated - but not remove it - to protect replay attacks

Thanks @ligi for your review
I have updated it

PR #1895, opened 2022-11-18
Create eip155-1229.json

Exzo-Network:
Adding Exzo Network chain.


ligi (changes requested):
The Icon exzonetwork does not exist - was used in eip155-1229.json

robledo6 (approved):

PR #1681, opened 2022-10-01
Add OctaSpace network

x9q:

PR #1913, opened 2022-11-21
Add OctaSpace network

x9q:
Hello,

Please add OctaSpace network, it's one another PR, previous one was closed #1681

Thanks


ligi (changes requested):
The Icon octa does not exist - was used in eip155-800001.json

x9q:
@ligi Changed 'octa' to 'octaspace' in eip155-800001.json

PR #1914, opened 2022-11-22
debank icon & info

lhypj:


ligi (changes requested):
width in json (../_data/icons/debank.json) is 1000 but actually is in imageDownload 400

ligi (approved):

PR #1931, opened 2022-11-24
create eip155-877.json

Dexit-Network:
could you please add RPC for Dexit Network


ligi (changes requested):
The Icon Dexit does not exist - was used in eip155-877.json

robledo6 (approved):

PR #1941, opened 2022-11-25
register μNode Testnet

yeti-detective:
Hi, I'm just wanting to register this testnet. There are not faucets at this time (though one or more will be added in the near future), and I'm not ready to expose the RPC, but I'd like to reserve the chain id because I have already put too much thought into it.
Let me know if any further work is needed on the PR

### DIFF
--- a/_data/chains/eip155-956.json
+++ b/_data/chains/eip155-956.json
@@ -1,0 +1,16 @@
+{
+    "name": "μNode Testnet",
+    "chain": "munode",
+    "rpc": [],
+    "faucets": [],
+    "nativeCurrency": {
+      "name": "Ether",
+      "symbol": "ETH",
+      "decimals": 18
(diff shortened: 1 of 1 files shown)


samuelmonday6:
I Need it now

yeti-detective:
> I Need it now

@samuelmonday6 was there something I can provide?

ligi:
The problem is the name - would be the first chain using this char and this not in the regex - ideally do a rename or make a very good point to extend the charset allowance

yeti-detective:
Ah yeah, I thought it was probably 50/50 whether that char would already be supported. No problem at all with renaming it.

ligi (changes requested):

also rebase to master

yeti-detective:
@ligi rebased & prettier-ized

PR #1949, opened 2022-11-27
Create eip155-493.json

ProofOfApes:
Proof of Apes Mainnet Addition


ligi (changes requested):
The Icon poa does not exist - was used in eip155-493.json

PR #1958, opened 2022-11-29
Update eip155-22052002.json

excelon-team:
Added a new RPC url with https format


excelon-team:
Thank you Ligi,

Do you know why the logo does not show up on chainlist.org?

Is there something we should fix?

Costis

------ Original Message ------

PR #1963, opened 2022-11-30
Add ZetaChain chains

CharlieMc0:
Adds the live ZetaChain Athens testnet chain and reserves the unreleased ZetaChain Mainnet chain.  

ZetaChain Athens Testnet - status: `active`

ZetaChain Mainet status - status: `incubating` ( I couldn't find the definition of `incubating` so I hope this was the right status for an unreleased chain)

Ran `npx prettier --write _data/*/*.json` so it should meet the style requirements.


ligi (changes requested):

CharlieMc0:
@ligi binary icon added into iconDownload. Anything else I need to change?

_data/chains/eip155-7001.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "ZetaChain Athens Testnet",
+  "chain": "ZetaChain",
+  "icon": "Zetachain",
```
ligi, line 4:
```suggestion
  "icon": "zetachain",
```

ligi (changes requested):

PR #1972, opened 2022-12-02
Add aves main-net network | Fixed 2

Toni-d-e-v:
I fixed all of those things, sorry for the times wasted @ligi also I'm in idena :)


ligi (approved):

ligi:
nice! we should also add Idena here - but as it is very different it needs a namespace definition first

PR #2004, opened 2022-12-11
Create eip155-1229.json, Create exzo,json - Add support for Exzo Network Mainnet

Exzo-Network:
Add support for Exzo Network mainnet and add the logo for the network.


ligi (changes requested):

Exzo-Network:
Sorry about that I overlooked that and just updated it to xzo. Thank you for all of your time as well!

ligi:
please remove the filename extension in iconDownload

Exzo-Network:
Removed the .png extension on the iconDownload file QmeYpc2JfEsHa2Bh11SKRx3sgDtMeg6T8KpXNLepBEKnbJ. Thank you.

PR #2017, opened 2022-12-13
Update p12 chain native token symbol

noyyyy:


ligi (changes requested):
Please provide a reasoning

noyyyy:
It's due to a mistake in the last pr https://github.com/ethereum-lists/chains/pull/2010. Our Team(P12 Team) decided to name it hP2 but I forget it. They remind me several minutes ago. Specifically, P12 Token would be an ERC20 on the production network such as Ethereum, and BNB Chain. P12 Chain would be a game-specific side chain. We name the native token as hP2 rather than hP12 and just want to distinguish it.

Btw, this chain currently used internally in the p12 team and there a time before it's broadcast to the public. So it doesn't matter if changing the token name and symbol.

ligi (approved):

PR #2019, opened 2022-12-13
add alyx chain icon

alyx-group:


ligi (approved):

alyx-group:
Halo there, may I ask how long it will take to show the new added icon in the [chain](https://chainlist.org/) site?

PR #2021, opened 2022-12-13
xanachain info added

SalmanNoborderz:


ligi (changes requested):
Icon URL must be IPFS CID link

SalmanNoborderz:
> Icon URL must be IPFS CID link

Sure I'll do that, what about the chainId as I can see that a chain with '8888' is already added?

SalmanNoborderz:
@ligi please check if the URL is correct now

ligi:
filename chainID does not match json

SalmanNoborderz:
> filename chainID does not match json

@ligi  There is already a json file with chainId 8888 in its name, how can I resolve that?

ligi:
https://github.com/ethereum-lists/chains#collision-management

SalmanNoborderz:
> https://github.com/ethereum-lists/chains#collision-management

@ligi so how should I add info for XANAChain here, do I edit the otherone? otherwise the filename and chainId won't match

PR #2026, opened 2022-12-13
Add BOMB Mainnet + Testnet: ChainID 2300 and 2399

cryptoaaron:
Included are spec files for both BOMB Chain id 2300 and BOMB Chain Testnet id 2399.  Also correct file specifying image has been included.


ligi (changes requested):

cryptoaaron:
I have added the file exactly as you requested.  Let me know if anything else requires changing 👍

cryptoaaron:
I see the problem causing build to fail and just pushed another commit with fix

cryptoaaron:
Sorry I closed the PR by accident while submitting prior comment, I just re-opened

ligi (approved):

PR #2027, opened 2022-12-13
Update chain 427 and 859

rampenke:
Fixed suggested issues


ligi (changes requested):
short name must be unique - but found `ZEETH` more than once

ligi (approved):

PR #1844, opened 2022-11-07
Add Firechain Mainnet & Rinia Testnet

7bp:

PR #2038, opened 2022-12-16
Change chain and network id for Firechain Main & Rinia Testnet

7bp:
Previously, I requested this add (https://github.com/ethereum-lists/chains/pull/1844) and now I kindly ask to make this change.


ligi (changes requested):
You cannot delete chains (replay attack problem) - only deprecate and add

7bp:
> You cannot delete chains (replay attack problem) - only deprecate and add

Thanks, I added back the old configs and added a status field.

ligi:
names can be changed to e.g. _old - no replay attack here

7bp:
I updated the names and short names to reflect the deprecate status.

PR #2040, opened 2022-12-16
Tfc patch 1

ngocyen1402:
Có


ngocyen1402:
> There

https://github.com/ethereum-lists/chains/pull/2040#issue-1500864377

ligi (changes requested):
change is empty -> no merge

ligi (changes requested):
There is no change in your PR

PR #2053, opened 2022-12-21
Update chains for Oort (formerly Computecoin)

metav:
Change the chain names after Computecoin rebranded to Oort. Add more chains for Oort.


ligi (changes requested):

ligi (approved):

PR #2056, opened 2022-12-21
add Vention Testnet

solidityx:


ligi (changes requested):

also:  The Icon vention does not exist - was used in eip155-741.json

PR #2070, opened 2022-12-26
add smarthost chain

smarthost-tr:
add smarthost logo


ligi (changes requested):
Please isolate the changes & do not do the changes to the README - IMHO these are not improvements

PR #2075, opened 2022-12-28
Solidityx patch 1 Vention

solidityx:


ligi (changes requested):
short name must be unique - but found `VSC` more than once

ligi (changes requested):
✓ ~/Downloads identify QmcNepHmbmHW1BZYM3MFqJW4awwhmDqhUPRXXmRnXwg1U4   14:11:43
identify-im6.q16: no decode delegate for this image format `' @ error/constitute.c/ReadImage/560.

solidityx:
> short name must be unique - but found `VSC` more than once

I've fixed that

ligi (approved):

PR #2082, opened 2023-01-01
Add BRNKC Chain

BearNetwork-BRNKC:
Add BRNKC Chain and BRNKC icons and BRNKC iconsDownload


_data/chains/eip155-641230.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "Bear Network Chain Mainnet",
+  "chain": "BRNKC",
+  "rpc": [
+    "https://brnkc-mainnet.bearnetwork.net",
```
ligi, line 5:
```suggestion
    "https://brnkc-mainnet.bearnetwork.net"
```

ligi (changes requested):

ligi (changes requested):

also the image has a problem - it is 0 bytes

PR #2085, opened 2023-01-02
XANAChain

SalmanNoborderz:


ligi (changes requested):
Reusing a chainID is problematic and only done in rare cases.
You will for sure need to add `"redFlags": ["reusedChainId"]`
And also show that the old one is not coming back.

SalmanNoborderz:
@ligi the old chainId '8888' is of AmbrosChain Testnet of whom nothing works anymore, this is of XANAChain Mainnet, what proof will be required that it is not coming back?

ligi:
e.g. a statement of them that they discontinued the network
but you will also need to add: "redFlags": ["reusedChainId"]

SalmanNoborderz:
> e.g. a statement of them that they discontinued the network but you will also need to add: "redFlags": ["reusedChainId"]

@ligi nothing of theirs work anymore(RPC, explorers, twitter, telegram), their website is static with no links working, no links on CMC are working also...

I'll add "redFlags": ["reusedChainId"]

_data/chains/eip155-8888.json
```diff
@@ -1,22 +1,25 @@
...
   "chainId": 8888,
   "networkId": 8888,
+  "icon": "xeta",
   "explorers": [
     {
-      "name": "Ambros Chain Explorer",
-      "url": "https://testnet.ambrosscan.com",
-      "standard": "none"
+      "name": "XANAChain",
+      "url": "https://xanachain.xana.net/",
```
ligi, line 20:
```suggestion
      "url": "https://xanachain.xana.net",
```

PR #2094, opened 2023-01-05
Create eip155-8098.json

devstreamux:
adding new streamux blockchain


devstreamux:
@ligi can you check

ligi (changes requested):

also:  Native currency symbol must have less than 7 chars

PR #2107, opened 2023-01-08
Added 1971 Atelier test network

dojoidiot:
Requesting addition of this network.


ligi (changes requested):

also add the faucets field again - can be an empty array - but must be present

ligi (approved):

PR #2126, opened 2023-01-11
Update to Tres Leches Blockchain [New RPC, New Website]

bladepool:
Added new RPC,
Changed website.
Updated Currency Name for simplification.


ligi (changes requested):

also the new RPC does not seem to work - so maybe - not add it?

bladepool:
> Please run the prettier to format your json according to the style [defined here ](https://github.com/ethereum-lists/chains/blob/master/.prettierrc.json) e.g. run `npx prettier --write _data/*/*.json`
> 
> also the new RPC does not seem to work - so maybe - not add it?

fixed so fixed rpc address as it was mispelled.

bladepool (commented):
updated

PR #2136, opened 2023-01-14
Update Fusion Network rpc, logo, config

fanbsb:
Fusion Network config update
1. RPC update.
2. Explorer update.
3. Icon update.
4. Testnet Network update.


ligi (changes requested):

fanbsb:
Thanks for the reminder, I have uploaded the icon files into the iconDownload folder, please review

ligi (changes requested):
: width in json (../_data/icons/fusion.json) is 1000 but actually is in imageDownload 31

fanbsb:
> : width in json (../_data/icons/fusion.json) is 1000 but actually is in imageDownload 31

Update the size for both 2 images

ligi (approved):

PR #2140, opened 2023-01-15
add BeOne Chain Testnet

beonechain:


duekz008:
Please review it, sir; we have already made a change as per your instruction

duekz008:
@ligi

ligi (changes requested):

PR #2145, opened 2023-01-16
Add Bitkubchain Testnet

ppanya:
# Added Bitkub Chain Testnet
  - Network ID: 25925


macvts:
@ligi we split the case to Mainnet #1629 and Testnet #2145

ligi (approved):

PR #2146, opened 2023-01-16
Change BKC Testnet Shortname

ppanya:
- change it, because it's conflict with the Mainnet one
  - BKC Testnet shortName will use "bkct" instead

### DIFF
--- a/_data/chains/eip155-25925.json
+++ b/_data/chains/eip155-25925.json
@@ -13,12 +13,12 @@
     "decimals": 18
   },
   "infoURL": "https://www.bitkubchain.com/",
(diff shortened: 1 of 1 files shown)


macvts:
@ligi  kindly request for your review, there is some conflict on #2145 that need to be change in case of the mainnet integration to run through.

ligi (approved):

macvts (commented):
@ppanya 
can we adjust from bkcscan-testnet to Bitkub Chain Testnet Explorer? maybe open another PR

robledo6 (approved):

PR #2151, opened 2023-01-17
Adding Gitshock Chain

avenbreaks:


ligi (changes requested):
Remove [httpsloader/bin/main/org/ethereum/lists/chains/https/Chain.kt](https://github.com/ethereum-lists/chains/pull/2151/files#diff-9e9761b0fde0374bd7f0d13e1197c5277b284c78a4d48d5ea5da6f5e90e7819f)

ligi (changes requested):
also:
 width in json (../_data/icons/gitshockchain.json) is 512 but actually is in imageDownload 1251

ligi (approved):

PR #2155, opened 2023-01-18
Update

heyoka-448:
Added Credit Smartchain Main Net


ligi (changes requested):
The explorer cannot be reached
Also why 2 json files for the same content?

heyoka-448:
Thanks for the feedback we will resolved it, the logo file is the same for asset and explorer.

ligi (approved):

PR #2161, opened 2023-01-18
Fix Etica Logo ipfs

kevin-wad:
Update etica.json

### DIFF
--- a/_data/icons/etica.json
+++ b/_data/icons/etica.json
@@ -1,6 +1,6 @@
 [
   {
-    "url": "ipfs://QmYSyhUqm6ArWyALBe3G64823ZpEUmFdkzKZ93hUUhNKgU",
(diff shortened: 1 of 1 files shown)


ligi (changes requested):
The content of the CID is not an image but a directory

kevin-wad:
apparently my ipfs uploader uploads only this way, can't find a way to upload just the image. If you have an ipfs uploader that uploads doesn't upload files into folders please let me know. Or if you know maybe you can upload the image and share the ipfs link, it would be very appreciated. Thank you very much for help and support

kevin-wad:
I found a solution, thanks!

kevin-wad:
> The content of the CID is not an image but a directory

I just realised that you had not given me a random Qm hash but the right ipfs hash of the logo png. Do you know why the logo is not showing up in the chainlist.org page although it's correct ipfs hash? Should it be set on another field than icon: to show on main page? Thanks

PR #2162, opened 2023-01-19
BLGChain Testnet Adding

waliair:
BLGChain Testnet Adding.


ligi (changes requested):
The Icon BLG does not exist - was used in eip155-123321.json

ligi (changes requested):

ligi (changes requested):

ligi:
prettier still fails on the icon

13364174888 (approved):

ligi (approved):

PR #2169, opened 2023-01-21
Update eip155-500.json

HyperonChainOfficial:
Added HyperonChain Information

### DIFF
--- a/_data/chains/eip155-500.json
+++ b/_data/chains/eip155-500.json
@@ -21,3 +21,37 @@
     }
   ]
 }
+
+{
+  "name": "HyperonChain MainNet",
+  "chain": "HPN",
+  "icon": "hyperonchain",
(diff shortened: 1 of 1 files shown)


ligi (changes requested):
This is not valid JSON & seem your ChainID is already taken

HyperonChainOfficial:
@ligi 
We are KYCed project from InterFi, PinkSale, CertiK(while onboarding), BitMart, CoinBits and more 3 CEX which is listing us in Feb 2023.

Our official partners are also good companies.

Even if 500 network id is taken from other blockchain we are one of the deserving candidate to get it. I am not saying remove them but better if you will keep provision to allow both of us to list on chainlist.

If you want we will share official KYC proof with above mentioned agencies/exchange

ligi:
No - will not do this - also do not care for any KYC - duplicate chainIDs are a risk to users - it allows replay attacks

PR #2171, opened 2023-01-21
add a TMY Chain to chainlist

viknsagit:


ligi (changes requested):
I could not resolve your CID with `ipfs get` in a reasonable time - and before having a back and forth here I require you to put it in iconDownload. Before this back and forth did cost me a lot of time that I do not have.

also your exporer dose not seem to work:
image

PR #2207, opened 2023-01-31
update explorer url

salim-mahat:


salim-mahat:
How to fix this error that appears while adding this network to Metamask.
code: -32602, message: "Expected null or array with at least one valid
string HTTPS URL 'blockExplorerUrl'. Received: ", stack: '{\n  "code":
-32602,\n  "message": "Expected null
or…fbeogaeaoehlefnkodbefgpgknn/common-6.js:19:36310)'

PR #2208, opened 2023-01-31
add zhejiang chain

barnabasbusa:
Adding public withdrawal testnet Zhejiang.

More info at: https://zhejiang.ethpandaops.io


ligi (changes requested):

PR #2226, opened 2023-02-02
Added Consta Test Network.

constaChain:


ligi (changes requested):
Cannot reach the explorer

constaChain:
> Cannot reach the explorer

_data/chains/eip155-371.json file updated.

ligi (approved):

PR #2230, opened 2023-02-02
Fix AVES icon

Toni-d-e-v:


ligi (changes requested):
Found file QmeXJHSmpCMjKhWqnKoeuCMy94B87nAZtjQvBo7WGjHHjb.png in ../_data/iconsDownload that is not referenced

PR #2241, opened 2023-02-05
Add new network

h-nashida:
New network adding Bandai Namco Research Verse Mainnet,
https://www.bandainamco-mirai.com/en/
chainId: 876


ligi (changes requested):
explorer is not reachable

h-nashida:
Thanks for the confirmation.
I mistakenly thought that I had to register the chain-id first and then build the blockchain network.
We have built the blockchain network, so please check again.
https://explorer.main.oasvrs.bnken.net

PR #2246, opened 2023-02-06
Add Stratos testnet and mainnet reservation

BoThe1K:


ligi:
if it is just a reservation - maybe add status incubating

BoThe1K:
@ligi testnet is operating, but we want also to reserve a chain id for mainnet. Should I do this separatelly?

BoThe1K:
Ok let me make clear and split chains

PR #2254, opened 2023-02-09
Update Scroll Alpha testnet

Thegaram:


ligi (changes requested):
I cannot reach https://blockscout.scroll.io

Thegaram:
> I cannot reach https://blockscout.scroll.io

Yes, this testnet will launch in a few weeks, but we hoped we could update the values here before that.

ligi:
I cannot validate EIP3091 compatibility
So the chain is not yet active? Then pleas add status incubating

Thegaram:
> I cannot validate EIP3091 compatibility So the chain is not yet active? Then pleas add status incubating

Added. Appreciate your help with this.

ligi (approved):

PR #2264, opened 2023-02-10
Add AirDAO Mainnet & Testnet networks

nacknime-official:


ligi (changes requested):
short name must be unique - but found `AIRDAO` more than once

ligi (approved):

PR #2276, opened 2023-02-13
Update BitTorrent Chain Mainnet explorer url

jiro-ono:


ligi (changes requested):
Pleas add it as alternative explorer - not replace it

jiro-ono:
@ligi just added commit to add it alternative explorer

ligi:
please add it after the other one - as the other one seems to live on the infoURL domain it IMHO should come first

jiro-ono:
@ligi prob better for the bttcscan one to be the main explorer used tho, since it's the etherscan version and better/preferred explorer of the 2

ligi:
would then be good to have someone from bittorrentchain.io chiming in

PR #2280, opened 2023-02-15
Update Polygon zkEVM testnet (1422)

Diegoescalonaro:
Updated Polygon zkEVM Testnet according to official docs:
- RPC URL: https://rpc.public.zkevm-test.net
- Chain ID: 1422
- Short name: testnet-zkEVM-mango

Renamed `eip155-1402.json` to `eip155-1422.json`


ligi (changes requested):
Please do not rename/delete  - you have to deprecate the old one and create a new one. Otherwise there can be replay attacks.

Diegoescalonaro (commented):
Ready! 🚀

ligi (approved):

PR #2290, opened 2023-02-16
Add ARC Blockchain mainnet and testnet

wombatlabs:


ligi (changes requested):

also shortName must be unique

PR #2302, opened 2023-02-19
Add eip155-96.json (Bitkub chain)

tanakorn0314:
I have added "Bitkub chain" to the list. Please consider adding it to the mainstream.

Bitkub chain is an EVM-compatible chain operated by the biggest crypto currency exchange platform in Thailand. It has potential to grow blockchain adoption in South East Asia.

Getting listed on Chainlist.org dramatically helps in user adoption.


ligi:
dup of #1629

PR #2309, opened 2023-02-21
add ferrum testnet

3ierratango:
Add ferrum testnet to list


ligi (changes requested):
explorer needs the standards field - can be "none"

ligi (approved):

PR #2313, opened 2023-02-22
add arcology testnet

sameplacewei:
please add arcology network into chainlist


ligi (changes requested):

ligi (approved):

PR #2324, opened 2023-02-24
add: canto testnet

jessepinkman9900:


ligi (changes requested):
short name must be unique - but found `CANTO` more than once

ligi (changes requested):
short name must be unique - but found `CANTO` more than once

jessepinkman9900:
@ligi bump

ligi:
@jessepinkman9900 short name is still not unique ..

jessepinkman9900:
@ligi my bad i thought it symbol was the issue. have fixed short name as well

ligi (changes requested):
still collides with eip155-740.json

jnsdls:
according to the CANTO docs they're "migrating" to this new chain: https://docs.canto.io/evm-development/testnet

@ligi in such a case is my assumption correct that the correct approach would be to:
- create a new shortname for this new entry (unanbiguous)
- mark the old chain (eip155-740 in this case) as `deprecated` status

submit a PR that does both on one go?

If so maybe @jessepinkman9900 can make this change in this PR, otherwise I'm happy to submit one today that makes these changes.

jessepinkman9900:
@ligi  @jnsdls 
have updated 740 as deprecated & fixed 7701 short-name

PR #2328, opened 2023-02-24
Added chain for Arévia.

countessellis:


ligi (changes requested):
shortName must match pattern "^[A-Za-z0-9-_]{1,64}$"

ligi (approved):

PR #2331, opened 2023-02-25
Create odin

ODINZYB:


ligi (changes requested):
It's for sure in the wrong folder - also infoURL looks wrong

Kywaldos (approved):

PR #2332, opened 2023-02-26
deprecated eip155-207.json

SiriusnetInstitute:


ligi (changes requested):
name must be unique - but found `SIRIUSNET` more than once

also add "status": "deprecated" to the deprecated chain

warewar1:
Please check angin!

ligi (changes requested):
there is no change since my review - please do not waste my time

PR #2349, opened 2023-03-02
submit doschain testnet to chainlist

Dien-DOS:


ligi (changes requested):
NativeCurrencySymbolMustHaveLessThan7Chars: Native currency symbol must have less than 7 chars

Dien-DOS:
I have changed the character, please check again for me.

ligi (changes requested):
short name must be unique - but found `DOS` more than once
also fix the other CI errors

PR #2388, opened 2023-03-09
Changed ChainID from 75000 to 75001

5050prop:


ligi (changes requested):
sorry to protect from replay attacks you cannot do this - but u can deprecate one chain and add another.

5050prop:
> sorry to protect from replay attacks you cannot do this - but u can deprecate one chain and add another.

this network is mine and @SsNiPeR1, old chain was deprecated by us because we didn't have any archive nodes, so we created a new chain with a different id.

5050prop:
the old resincoin with chainid 75000 is no longer supported

ligi:
you need to deprecate it in this PR and not just delete it

PR #2397, opened 2023-03-11
Update ohoscan.json

ohonetwork:
Logo doesn't show on chainlist.org, so try changing to SVG format.


Oluwarichee:
I love you forever my lord I accept Christ as my wife and only my heart and
soul and my second one to share my thought with I will never change my
words oh lord 👏💡👑🫅

PR #2404, opened 2023-03-11
Update oho.json

ohonetwork:
Logo doesn't show up at https://chainlist.org, so trying SVG format.


ligi (changes requested):
please add the other image - not replace it - also report the bug at chainlist.org

ohonetwork:
> please add the other image - not replace it - also report the bug at chainlist.org

Thank you for your suggestion. I will report to chainlist first before changing it to SVG. Thanks

PR #2405, opened 2023-03-11
Remove ohoscan.json

ohonetwork:
Removing this file as it's no longer needed in eip155-39815.json.


ligi:
combine it with the other PR

PR #2408, opened 2023-03-13
Add okbchain testnet and mainnet

ilovers:
The testnet and mainnet are not yet online, we just want to keep the chain id first.


ligi (changes requested):
RPC chainId (65) does not match chainId from json (195)

ilovers:
> RPC chainId (65) does not match chainId from json (195)
Updated. Our new chain is not launched and i have added the status to incubating in the json file.  I will update these json files when the chain is launched. Thanks!

PR #2412, opened 2023-03-13
Add fibonacci icon file

FiboChain:
Add fibonacci icon file


ligi (changes requested):
no change -> no merge

PR #2432, opened 2023-03-16
Add eip-8387: Dracones mainnet

farcaller:
nb: the rpc url is having issues and that's expected. The url *is* correct.

### DIFF
--- a/_data/chains/eip155-8387.json
+++ b/_data/chains/eip155-8387.json
@@ -1,0 +1,19 @@
+{
+  "name": "Dracones Financial Services",
+  "title": "Dracones Financial Services (a joke mainnet for the Wolfery MUCK)",
+  "chain": "FUCK",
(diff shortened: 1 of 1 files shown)


ligi (changes requested):
will only merge a joke if it is funny - so far it is not ..

farcaller:
> will only merge a joke if it is funny - so far it is not ..

Right, that's fair.

To elaborate on it: It's a joke in a sense that while the network is public (or will be public once I fix the dumb ssl trust issue), its value is effectively tied to the single website so it's not of much use as a general compute network. Still, it's functional as one.

I can remove the joke part out of the description if that helps. The plan is to keep the network operating as a mainnet with similar approach to stability (of course it being a volunteer-running extension to a MUCK makes its limits on the computational power and replication).

PR #2443, opened 2023-03-17
Add the image PR check

farcaller:
This adds a little automation to verify the basic image parameters if the PR adds or modifies icons.

Checks:
* ipfs should load (through cloudflare)
* width and height must match
* type should be one of png,jpg,svg and must match

Example of a good PR: https://github.com/farcaller/chains/pull/2
Example of a failing PR: https://github.com/farcaller/chains/pull/1

It is a slightly flaky test by nature as IPFS load might fail.

### DIFF
--- a/.github/workflows/pr_checks.yml
+++ b/.github/workflows/pr_checks.yml
@@ -1,0 +1,99 @@
+name: Run PR Checks
+
+on:
+  pull_request:
+
+jobs:
+  validate_icons:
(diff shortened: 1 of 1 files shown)


ligi:
found one edge-case - when icon is supplied in iconDownload it should not fail: https://github.com/ethereum-lists/chains/pull/2457

jonesk7734 (approved):
👍

PR #2446, opened 2023-03-18
Add a slightly better linter check

farcaller:
Reports the diff with changes instead of asking the user to run the tool: [example failure](https://github.com/farcaller/chains/actions/runs/4456330928/jobs/7826703319?pr=3).

It will also run on only the changed files since the previous check run.

Updated the original check to run only on the commits to master.


jonesk7734 (approved):
Ok

PR #2474, opened 2023-03-22
Submit Dexilla chain

0xSheldon:
Dexilla Chain is a Layer 2 solution that is built on top of the Optimism protocol.


ligi:
You might want to have a parent field then to point to the corresponding L1

0xSheldon:
> You might want to have a parent field then to point to the corresponding L1

Update parent chain already.

ligi (approved):

PR #2485, opened 2023-03-24
Add superchain testnet and mainnet

superexspace:
Add superchain testnet and mainnet


ligi (changes requested):
RPC chainId (1969) does not match chainId from json (1970)

superexspace:
> RPC chainId (1969) does not match chainId from json (1970)

Fixed.

ligi (changes requested):
the Icon Super does not exist - was used in eip155-1969.json

superexspace:
> the Icon Super does not exist - was used in eip155-1969.json

Fixed

ligi (changes requested):

also the explorers do not seem to be explorers - at least not  seem to be EIP3091 compliant

PR #2501, opened 2023-03-25
update icon doschain

Dien-DOS:


ligi (approved):

Dien-DOS:
hello, why icon my chain not display in in chainlist?

PR #2534, opened 2023-03-29
Qitmeer related network

bCoder778:


ligi:
are these incubating chains?

bCoder778:
Yes, these are all Qitmeer-related incubation chains

bCoder778:
@ligi Yes, these are all Qitmeer-related incubation chains

ligi:
then please add the field "status":"incubating"

PR #2547, opened 2023-03-30
add combo testnet

combo-labs:
add combo testnet


ligi (changes requested):
The explorer is not EIP3091 compatible when it comes to transactions

tribeless88 (approved):

PR #2618, opened 2023-04-11
Create eip155-67588.json

dallinandteam:
Register Cosmic Chain Testnet to Chainlist


cosmicchainofc:
error with change name?
submit new request?

ligi (changes requested):

PR #2635, opened 2023-04-13
Sonarqube

m-eld:


ligi (changes requested):
Please elaborate - the sonacube add does not make sense to me

PR #2663, opened 2023-04-18
Add Zafic icon

TuanTerTH:


ligi (changes requested):
I cannot validate EIP3091 compatibility - please provide me with deeplinks so I can do

TuanTerTH:
> I cannot validate EIP3091 compatibility - please provide me with deeplinks so I can do

I fixed it, Please check.

PR #2666, opened 2023-04-18
Changed Neonlink Icon format

neonlinkdev:


ligi:
why?

ligi:
I think svg is actually better and also supported by e.g. https://chainid.network - but seems there seems to be a problem with this certain SVG - but I think instead of rasterizing the bug should be fixed

PR #2672, opened 2023-04-19
Create eip155-142857.json

David0817:
Add ICPlaza Mainnet


ligi (changes requested):
Exception in thread "main" java.lang.IllegalStateException: The Icon icplaza does not exist - was used in eip155-142857.json

David0817:
close

PR #2673, opened 2023-04-19
Create icplaza.json

David0817:
Add icplaza Icon


ligi:
combine with the other PR

PR #2685, opened 2023-04-20
Add gather icon testnet

vinodsharmak:


ligi:
can you elaborate on the change?

ligi:
should be fixed in the frontends then - jpg is not good for icons

PR #2687, opened 2023-04-20
add satoshichain testnet

dglw:
add satoshichain testnet to chainlist


ligi (changes requested):
the CID is not an image

dglw:
CID changed with new one

ligi (approved):

PR #2691, opened 2023-04-21
Added 5ireChain Thunder

neel-ds:
added 5ireChain Thunder Testnet


ligi (changes requested):
explorer cannot be reached

neel-ds:
> explorer cannot be reached

Changed the explorer url

ligi (changes requested):
Explorer does not seem to be EIP3091 compatible - change to `ǹone`

PR #2759, opened 2023-05-05
add structx

trevatk:


ligi (changes requested):
icons do not fail - formatting does:

trevatk:
> 

all pretty now

ligi (approved):

PR #2809, opened 2023-05-11
Create eip155-274.json

mauricio-b-r:


ligi (changes requested):
explorer looks broken - so cannot check EIP3091 compliance

mauricio-b-r:
> explorer looks broken - so cannot check EIP3091 compliance

We are working on it, I'll trigger the re-request when we finish fixing the explorer

ligi (approved):

PR #2807, opened 2023-05-11
update: rpc;

oliviachef:

PR #2811, opened 2023-05-11
RevertN°1 "update: rpc;"

romanR1H2:
Reverts ethereum-lists/chains#2807


oliviachef:
does not makes sense;

PR #2837, opened 2023-05-15
added pepechaint-testnet details

pepem00n:


deep709:
Thanks

pepem00n:
tried multiple times to upload the icons to ipfs, all the time locally we can resolve the icon but the merge checks fails, can we get more info on how to properly do this?

thanks

ligi:
Please just do something else - it is clear you should not run a chain ..

pepem00n:
> 

thanks for the unsolicited advice.

i'm not a technical person behind this project and for me the CID provided in the json is returning the image with the properties in the json, so if you could help me with some actual advice instead of sarcasm that would be great.

PR #2862, opened 2023-05-18
Add Rollux Mainnet and Update Rollux Testnet info

fernando-syslabs:
Thanks in advance!


ligi (changes requested):
I cannot reach https://explorer.rollux.com

fernando-syslabs:
@ligi, we have the explorer in better shape now. Can you try again?

PR #2869, opened 2023-05-19
Adding Dragonfly Mainnet and Mantis Testnet

RobertOttBS:
Added chaindata and icons for Dragonfly mainnet and Mantis testnet


RobertOttBS:
improved format and changed icons to png to pass tests with width and height

PR #2879, opened 2023-05-22
Add Scroll Sepolia (534351) & update other Scroll chains statuses

dghelm:
Established ChainId for in-development Scroll testnet on Sepolia.

Additionally, updates statuses to show that Scroll Alpha Testnet on Goerli is our only active network. PreAlpha has been deprecated, and mainnet is not yet active.


OrestTa:
Good to know there are statuses <3

ligi (approved):

PR #2887, opened 2023-05-23
Update domain for Super chain

superexchain:


ligi:
The old domain seems to still be active - why the change - do you have proof?

superexchain:
> The old domain seems to still be active - why the change - do you have proof?

Yes, I just closed the old domains, May I change the new domains, Thanks.

PR #2914, opened 2023-05-29
update siberium icon

siberium-net:


ligi:
why? SVG is better than PNG

siberium-net:
This one is newer version, .png because some apps still don't display .svg.

ligi:
that should be fixed in those apps - please stick to svg

PR #2928, opened 2023-05-31
Add LUKSO Mainnet and Testnet

frozeman:
This adds LUKSO Mainnet and Testnet to the chains list.

Note LUKSO Mainnet uses the same chainID as the deprecated and stopped Kovan Testnet.

LUKSO is the 3rd eth2 like mainnet (that we are aware of), started fully decentralized with [10336 validators keys from over 200 genesis validators](https://dune.com/hmc/lukso-genesis-validators) on May 23rd 4:20pm UTC (See: https://medium.com/lukso/genesis-validators-start-your-clients-fe01db8f3fba)

Stats pages:
- https://stats.execution.mainnet.lukso.network
- https://explorer.consensus.mainnet.lukso.network
- https://explorer.execution.mainnet.lukso.network

Twitter: https://twitter.com/lukso_io


ligi (approved):

ligi:
It is mergeable - but a bit disappointed you reused a chainID and expose your users to replay attacks just for the number and do not use a standard in your explorers. 
IMHO good frontends should filter these chains/explorers by default - and for redFlags display a big fat red warning.
Also you might want to have a PR to https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md - as there 42 as an example is still kovan

frozeman:
Thank you!
LUKSO started with 42M supply, at 4:20 UTC and 42 is everywhere :P
I think it is unlikely that someone that developed on kovan will use the same private key on LUKSO.
also we push UP contract accounts as the main way to interact, which prevent that issue in general.

But i agree, there is a slight risk

PR #3001, opened 2023-06-12
Changing icon to PNG instead of JPG for Living Assets Mainnet

zorjak:
For some reason JPG image that we uploaded before does not show on website. I would like to try does it work with PNG image


ligi:
Icon shows up fine

image

zorjak:
Thank you very much for confirming.

zorjak:
Hi @ligi ,

Sorry for bothering you, but if you can reply I would really appreciate your response.
Can you tell me when you normally promote new stuff to prod? I am curious because the icon is still not visible on the chainlist website, and I am wondering when that could be.

Thank you in advance for your reply.
Zoran

PR #3010, opened 2023-06-14
Add Zora Network

tbtstl:
This PR adds network information for Zora's network, an optimistic rollup chain for zora.co. 

More information is available at [zora.energy](https://zora.energy).

### DIFF
--- a/_data/chains/eip155-7777777.json
+++ b/_data/chains/eip155-7777777.json
@@ -1,0 +1,22 @@
+{
+  "name": "Zora",
+  "chain": "ETH",
+  "rpc": ["https://rpc.zora.co/"],
+  "faucets": [],
+  "nativeCurrency": {
+    "name": "Ether",
+    "symbol": "ETH",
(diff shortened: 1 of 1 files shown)


iainnash:
➕

moriowanibe (approved):

ligi (changes requested):

iainnash:
It looks like that prettier style is what is in this JSON: 
```
> npx prettier --check '_data/*/*.json'
Checking formatting...
All matched files use Prettier code style!
```

ligi:
this is the last report I got from the action - but also somehow the workflows for your PR are not running currently - maybe because of the many force pushes 
Maybe open as another PR - nothing I can do here

tbtstl:
> this is the last report I got from the action - but also somehow the workflows for your PR are not running currently - maybe because of the many force pushes 
> 
> Maybe open as another PR - nothing I can do here

Understood, will close this PR in favor of a new one.

PR #3015, opened 2023-06-15
Add CamDL Mainnet

sereyvathanatum:
CamDL is blockchain network run by Ministry of Economy of Finance in Cambodia, with the initiative to provide the public a platform to experiment or develop any Web3 application to solve everyday problems or be part of the DeFi ecosystem.


ligi (changes requested):
You need to add redFlags reusedChainId - and think twice if you really want to expose your users to replay attacks

sereyvathanatum:
Thanks Ligis, Thanks for reminding. We are fully aware about the risk of replay attack, but due to different in region Argentina (testnet) vs Cambodia. We dont think that replay attack has too much of an impact to our users. Best regards, Vathana

ligi (approved):

PR #3034, opened 2023-06-20
feat: zora network icon

kulkarohan:
This PR adds the icon for Zora's network, an optimistic rollup chain for zora.co.

More information is available at [zora.energy](https://zora.energy/).


kulkarohan:
> Icon should be referenced in the chain file 😅

oops lol, just added

PR #3038, opened 2023-06-21
Horizen Yuma Testnet deprecated

ddrvar:


ligi:
Do you have a link about this? Chain still seems to produce blocks

ddrvar:
@ligi it is deprecated, and is only for internal use
more about it can be found here https://blog.horizen.io/deprecation-of-dune-and-yuma-testnets/

PR #3082, opened 2023-06-29
Add eip155-534849

shinamaxi:
Created _data/chains/eip155-534849.json to add Shinarium network to chain list.


cynnadior (approved):

alokeutpal (approved):

ligi (changes requested):
https://shinascan.shinarium.org is not working

PR #3097, opened 2023-07-02
Update .prettierrc.json

sammyfilly:
Prettier isn’t compiling and working with my projects it requires an updated version


alokeutpal (approved):

ligi (changes requested):
why?

PR #3113, opened 2023-07-05
Updated Chain ID for ShimmerEVM Testnet

anistark:
Recently due to certain conditions this testnet chain was reset and resulted in a change in Chain ID.
Also, fixing a mistake on decimals from before.


ligi (changes requested):
You cannot just delete a chain - you need to deprecate the old one and add the new one

anistark:
> You cannot just delete a chain - you need to deprecate the old one and add the new one

Ah I see. How to deprecate a chain?

ligi:
"status" : "deprecated"

ligi (changes requested):
RPC chainId (1072) does not match chainId from json (1071)

you might need to remove the RPCs from the deprecated chains if you reuse the same addresses

ligi:
please also rename one of the shortnames - so they are unique

ligi (approved):

PR #3127, opened 2023-07-07
cifdaq testnet

Debopriya14:
cifdaq testnet


ligi (changes requested):
The Icon cifdaq does not exist - was used in eip155-30720231.json

PR #3139, opened 2023-07-10
Add Sepolia Public Goods Network

wbt:


wbt:
I see that validate_formatting failed, but can't tell why it failed or how to fix it.

ligi (changes requested):

wbt:
Done, but that requirement is quite unintuitive, especially when it requires correction to a whitespace pattern directly copied from the readme for this repo (example around bridges where linking to a parent).  
Prettier also changed every single JSON file under _data, apparently a line-ending difference which VSCode doesn't visualize as a diff. I didn't commit all of them.

ligi (approved):

PR #3153, opened 2023-07-12
Create eip155-424.json

Kweiss:
Adding PGN Network Mainnet details


ligi (changes requested):

Kweiss:
:doh:

thanks, fixed

ligi (approved):

PR #3181, opened 2023-07-17
Add chain-id for REAP

helloreap:
Main-net: 221230
Test-net: 221231


julioAk47 (approved):

minimayki6232 (approved):

ligi (changes requested):
short name must be unique - but found `REAP` more than once

ligi (approved):

PR #3217, opened 2023-07-24
add bitnet

masayoshikob:


masayoshikob:
hi @ligi i've fixed the links and removed the end slashes, but couldn't figure out the problem with the `validate_formatting` would you be able to point out what am i missing?

ligi (changes requested):

masayoshikob:
all done

ligi:
does not look this way

masayoshikob:
well, i've ran prettier, but now it's asking for the faucet link

masayoshikob:
i'll open another pr, this became too messy

PR #3279, opened 2023-08-04
Update Url's and Logo

heyoka-448:
We have consolidated our urls as subdomains from the main creditsmartchain.com we also updated the logo file to the most recent version.


ligi (changes requested):
Please rebase to master

PR #3281, opened 2023-08-04
feat(chain): adding new chain, beam testnet 13337

dfenstermaker:


ligi (changes requested):
The Icon beam does not exist - was used in eip155-13337.json

dfenstermaker:
@ligi thanks for taking a look, should be fixed now

ligi (approved):

PR #3291, opened 2023-08-07
Add Miexs Mainnet Network

chainkloud:
- Add Miexs Mainnet json
- Add Miexs Mainnet Icon


ligi (changes requested):
short name must be unique - but found `ALL` more than once

ligi (approved):

PR #3300, opened 2023-08-09
add DeBank Testnet

cryptocifer:


Qmig (commented):
Ok

ligi (changes requested):
rebase to master

ligi (changes requested):
also short name cannot be empty

ligi (changes requested):
name must be unique - but found `DEBANKTESTNET` more than once

ligi (changes requested):

ligi (changes requested):
name must be unique - but found `DEBANKTESTNET` more than once

Please do not submit chainIDs that you do not use - we cannot delete chains as we need to protect users from replay attacks

ligi (approved):

PR #3339, opened 2023-08-16
chore: add beam mainnet

RobbyUitbeijerse:
Configuration is based off the existing eip155-13337.json for Beam testnet


0xSilkweave:
Suggest adding chain icon and explorer icon (and JSON) as per example https://github.com/ethereum-lists/chains
Suggest changing shortname to something unique to resolve checks, as per EIP-3770 https://eips.ethereum.org/EIPS/eip-3770

RobbyUitbeijerse:
> Suggest adding chain icon and explorer icon (and JSON) as per example https://github.com/ethereum-lists/chains Suggest changing shortname to something unique to resolve checks, as per EIP-3770 https://eips.ethereum.org/EIPS/eip-3770

done & done!

RobbyUitbeijerse:
Updated the ipfs protocol url, missed that one and blatantly copied the URL from my browser when testing the icon.

@dfenstermaker his PR got merged before this one, but we want 'beam' as a shortname - 'Subnet' is not something we use in our communication or outward facing products - also it's Beam is marketed with a capitalised B, so it makes sense to use that name in the explorer name too 

lmk!

ligi (approved):

PR #3363, opened 2023-08-21
Add Krest network

neutrinoks:
Adding spec for Krest network with icon

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1205267245821359


neutrinoks:
Due to the github actions are checking for a matching EVM ChainID, we have do to our runtime-upgrade first. I hoped, it works the other way round... register an EVM ChainID, then fix it... Runtime-upgrade of our blockchain is scheduled for next week, so I will update this PR next week again :+1:

ligi:
having it the other way around would mean we have wrong data - I do not think this would be good.

neutrinoks:
> having it the other way around would mean we have wrong data - I do not think this would be good.

We will do our runtime upgrade on the beginning of next week, then the ID will fit and then I will update the last changes to get this finished :+1:

neutrinoks:
@ligi I pushed a little too early, because I was expecting that the GitHub-actions usually take much more time :see_no_evil: sorry!!! Can you initiate the GitHub-actions again? Should work now...

PR #3383, opened 2023-08-25
New Chains 78430, 78431, 78432

dfenstermaker:


ligi (changes requested):
why do you remove the beam icon?

dfenstermaker:
@ligi there isnt an icon there and this is the fastest way to get schema tests to pass

ligi:
your schema tests pass.
but there was an icon and you removed it ..

dfenstermaker:
@ligi ok can we merge it? or want me to put the icon back?

ligi:
so far I think removing the icon is a regression - so I would not merge it

dfenstermaker:
@ligi fixed

ligi (approved):

PR #3392, opened 2023-08-28
add_worldland_network

Hongjongnam:
hello.
I would like to register on the chain list. I'd really appreciate it if you could tell me in detail what I'm lacking.


Hongjongnam:
A fail occurred during the build process. Can you tell me what error it is?

ligi (changes requested):
The Icon worldland does not exist - was used in eip155-103.json

PR #3396, opened 2023-08-29
Change the xpla chain-ID to 3701->37. Delete seedCoin.

park1847:
I want to use the chainId 37 used by SeedCoin-Network as XPLA.

All sites related to SeedCoin-Network are inaccessible.
"infoURL": "https://www.seedcoin.network/"
"rpc": ["https://node.seedcoin.network"]

thank you


ligi (changes requested):
you cannot delete a chain - only mark it deprecated

PR #3411, opened 2023-08-30
Add Metachain Testnet

swarna1101:


swarna1101:
@jonesk7734 , @Quantumjumping , our Chain is still no visible on the website. Can you let me know when it will be put up there and can be used?

PR #3417, opened 2023-08-31
Fix Nautilus Triton testnet so it has a unique name

credfeto:
Changes the ``Nautilus Trition Testnet`` name from ``Nautilus Chain`` to ``Nautilus Trition Chain`` to avoid duplicate name


ligi (approved):

jonesk7734 (commented):
😀

PR #3424, opened 2023-09-02
Update website

FrederikBolding:
Update website to add persisted filters and filter out chains that don't have a valid RPC.


ligi (approved):

ligi:
Nice - thanks!

PR #3438, opened 2023-09-05
Add Mara testnet

blewater:


ligi (changes requested):
RPC chainId (123456) does not match chainId from json (1122)

PR #3446, opened 2023-09-06
Add Highbury Network

sudopunk:


sudopunk:
linted and updated

sudopunk:
@ligi The icon link seems to resolve fine but the asset file isn't passing validation. So I have updated the 'bafy' hash instead.

sudopunk:
@ligi I hope everything's in order. Please take it forward. Thanks.

sudopunk:
The IPFS hash is resolving as intended when using it with the link in the validation script.

ligi (approved):

PR #3452, opened 2023-09-07
new pr from vrcscan-mainnet vrcscan-testnet and for icon

virtualsole:


virtualsole:
Hi,
we provide details of our mainnet and testnet and also the icon of vrcscan
please add it in etherium chainlist

PR #3447, opened 2023-09-06
Add DeFiMetaChain Testnet ("Changi")

johnny-random:
Kindly add the provided chain. It is the upcoming EVM extension of [DeFiChain](https://defichain.com/).

PR #3468, opened 2023-09-08
Update DeFiMetaChain Testnet ("Changi")

PaulSimode:
Updating chain name, as "Testnet" was omitted. Original PR for adding was #3447.


johnny-random:
@ethereum-lists

Kindly have a look, seems like there's an error that's outside of my reach... :x

PR #3491, opened 2023-09-11
Add Bitfinity Devnet and Production chain ids

veeso:
We've added the devnet and production chain ids


ligi (changes requested):
explorers do not seem to work

veeso:
@ligi Yes, we're currently working on fixing the explorers, the PR is more likely to reserve these chain IDs which we'll be using soon. If we need to insert those after we have our production env and devnet running, then we can ping for the review when we're ready

ninooo123 (approved):

PR #3463, opened 2023-09-08
Deprecate eip155-444.json

aureliusbtc:
Deprecates Frenchain Testnet, currently using Chain ID 444, in preparation of a new PR for an unrelated testnet to use chain ID 444. 

RPC is fully down. 
```
cast block latest --rpc-url https://rpc-01tn.frenchain.app
Error:
error sending request for url (https://rpc-01tn.frenchain.app/): error trying to connect: dns error: failed to lookup address information: nodename nor servname provided, or not known
```

https://chainlist.org/chain/444

Explorer no longer exists (https://testnet.frenscan.io/), and redirects to https://app.tryethernal.com. 

Twitter account has been deleted: https://twitter.com/fren_chain, which is linked from the official site provided in the chain information: https://frenchain.app/

No tweets mention the Frenchain Testnet since November 2022.

PR #3492, opened 2023-09-11
Add Synapse Sepolia Testnet in place of deprecated chain

aureliusbtc:
In #3463, the testnet that is now down which was using chain ID 444 is now deprecated. 

Please let me know if this correctly adds new chain information in place of a deprecated chain, I could not find an example prior, except in the readme which mentions: "The only way to get a chain reassigned is when the old chain gets deprecated. This can e.g. be used for testnets that are short-lived."


ligi (changes requested):
you need to add `"redFlags": ["reusedChainId"]`

aureliusbtc:
Updated @ligi

jonesk7734 (approved):
😀

ligi (approved):

PR #3495, opened 2023-09-12
Create gold X logo

goldxchain:
Please add GOLDX ICON


ligi (changes requested):
please do it like every1 else - this does not work

PR #3522, opened 2023-09-18
Update eip155-14.json

fz22gq:
Add https://rpc.ftso.au/flare


fz22gq:
Please assist me in resolving the 4th check

PR #3643, opened 2023-10-07
add orderly mainnet

zlog-in:


ligi (changes requested):
Explorer does not seem to work:
image

zlog-in:
It is ok. Currently, there is no transaction on this mainnet because this mainnet is up since this friday (06.10.2023), we haven't deposit ETH to this L2. It should be active after we have set up bridge from Ethereum to this L2.

ligi (approved):

PR #3658, opened 2023-10-10
Alternative block explorer for a few chains

sneg55:
Added alternative block explorer


sneg55:
@ligi can you review this PR? Thanks!

PR #3677, opened 2023-10-15
Remove duplicate

AlicanC:
I've been trying to figure out some random issues I've been having and realized they were being caused by my end being case insensitive, and us having these two files with the same content: `lachain.json` and `LaChain.json`.

This PR removes `LaChain.json` which seems to be unnecessary.

### DIFF
--- a/_data/icons/LaChain.json
+++ b/_data/icons/LaChain.json
@@ -1,8 +1,0 @@
-[
-  {
-    "url": "ipfs://QmQxGA6rhuCQDXUueVcNvFRhMEWisyTmnF57TqL7h6k6cZ",
-    "width": 1280,
-    "height": 1280,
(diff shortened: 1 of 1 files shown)


ligi (changes requested):
The Icon LaChain does not exist - was used in eip155-274.json

AlicanC:
Oh, thanks.

So the file was indeed necessary.

I guess these files were the same not because of an unnecessary duplication, but another PR replacing contents of one with the other (probably because of a case sensitivity issue on the author's side):
https://github.com/ethereum-lists/chains/commit/487ea8c649998ab6333bc6d913660131f891e945#diff-2f4892204dbaa761361dd2851e5765e708b0bce2d13a8c2a4275b8da169d4c8b

PR #3678, opened 2023-10-16
update default explorer for eip155-534352.json

HAOYUatHZ:


ligi (changes requested):
Can u elaborate on the change? Both explorers are up and valid - why remove 1?

HAOYUatHZ:
> Can u elaborate on the change? Both explorers are up and valid - why remove 1?

(To clarify: I am from the scroll team, see https://github.com/scroll-tech/scroll/graphs/contributors)

we have both https://blockscout.scroll.io (using blockscout) and https://scrollscan.com/ (built by the etherscan team)
however, blockscout sometimes suffers from performance issues and consumes a lot of CPU resources.

if both https://blockscout.scroll.io and  https://scrollscan.com/ are in `_data/chains/eip155-534352.json`
and if users add scroll network via https://chainlist.org/
the default explorer will be https://blockscout.scroll.io
deleting the first entry, https://blockscout.scroll.io, can let the users use https://scrollscan.com/ by default

sorry for not clarifying it before

ligi:
can't you just switch the explorer positions instead of deleting one?

HAOYUatHZ:
> can't you just switch the explorer positions instead of deleting one?

that's also doable, updated, thx!

HAOYUatHZ:
@ligi can you take a look again? thx!

ligi (approved):

ligi (approved):

PR #3713, opened 2023-10-20
Update and rre

heyoka-448:


ligi (changes requested):
you cannot delete a chain - you can only mark it deprecated

PR #3722, opened 2023-10-22
add kafirchain-testnet

Kafiirr:


Kafiirr:
@ligi 
What should I do, where is the problem?

PR #3739, opened 2023-10-26
add cross_value_chain

0xAli0:


ligi (changes requested):
ChainVerse seems to be still up and running - so this is a no go - use another chainID

PR #3769, opened 2023-11-01
Add HYTOPIA Mainnet & HYTOPIA Testnet

iamarkdev:


macvts (changes requested):
ChainID need to be state differently, won't do for the same chainid (2911) to both mainnet and testnet

iamarkdev:
> ChainID need to be state differently, won't do for the same chainid (2911) to both mainnet and testnet

Ahh, apologies @macvts - resolved.

macvts (approved):

_data/icons/hytopia.json
```diff
@@ -0,0 +1,8 @@
+[
+  {
+    "url": "ipfs://QmcSJmAfk3iay6uCGJxgUPd9pTsRxbtF4BrNnC5txE9cg8",
+    "width": 400,
+    "height": 400,
+    "format": "png"
```
macvts, line 6: Format is wrong, cause failure to pr merging

macvts (changes requested):

PR #3800, opened 2023-11-13
rename xgon chain to X1 chain

ilovers:


ligi (changes requested):
Was the same chain renamed - do you have some pfoof of that?

giskook:
> Was the same chain renamed - do you have some pfoof of that?

Hi @ligi ,
Thanks for your question, you can refer to this pr: https://github.com/ethereum-lists/chains/pull/3771 we change okbc to xgon.  also both of @ilovers and I was verified by okx, I can find more colleagues to proof this if needed. 
Thanks in advance

ligi (approved):

PR #3801, opened 2023-11-13
Add_dxb_chain testnet

Osman-Mahmood:


ligi (changes requested):
you cannot just change a chain:

https://github.com/ethereum-lists/chains#collision-management

Osman-Mahmood:
The thing is i changed a chain and add my chain because there chain that already exists not show in evm chain list's that's why chain id 195 is our testnet and 71 id is for mainnet. both are not present in these links https://chainlist.org/.
The name with these files only exists but not actually present so please approve our request.I already test the ./gradlew run
 test and its good.The main purpose is to show logos of our coin.My pr test check is done and good. My third commit is good.
and its added in the main branch when it will show here?
https://chainlist.org/
https://chainlist.wtf/
Thanks and regards

Osman-Mahmood:
Screenshot 2023-11-13 at 5 22 27 PM
chain id 195 not present
Screenshot 2023-11-13 at 5 23 53 PM
chain id 71 not present

ligi:
you need to enable testnets in the filters 

image

PR #3848, opened 2023-11-24
Add Abey Mainnet.

thejasonhsu:


macvts (changes requested):
need to specify explorer standard like
"standard": "none" or "EIP3091"

PR #3863, opened 2023-11-27
Adding MaalChain testnet

maalchain:
Adding MaalChain testnet with chain ID 7860


maalchain:
Why is PR Check failing? Need help adding this testnet to Chainlist

PR #2279, opened 2023-02-15
Add Hedera networks

Neurone:

PR #3873, opened 2023-11-27
Updating HashScan's URLs

Neurone:
Now HashScan redirects URLs as requested by [@ligi's comment](https://github.com/ethereum-lists/chains/pull/2279#discussion_r1108699662) in #2279.

This PR updates the explorer's URLs accordingly and flags them as EIP3091 compliant.

### DIFF
--- a/_data/chains/eip155-295.json
+++ b/_data/chains/eip155-295.json
@@ -18,8 +18,8 @@
   "explorers": [
     {
       "name": "HashScan",
-      "url": "https://hashscan.io/mainnet/dashboard",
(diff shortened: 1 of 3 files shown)


Neurone:
You should be able to see accounts using the address endpoint like these:
- https://hashscan.io/mainnet/address/0x00000000000000000000000000000000003e014d
- https://hashscan.io/testnet/address/0x00000000000000000000000000000000005b5381

Please note in Hedera you need to create an account, or it's not existent by default.

PR #3891, opened 2023-11-30
Update Cronos Explorer Link in eip155-25.json

aw126:
Cronos mainnet new explorer has just been launched ([official announcement](https://blog.cronos.org/p/introducing-the-new-cronos-explorer-7ee)), updating the Cronos explorer link from Cronoscan to the new explorer: https://explorer.cronos.org/.

### DIFF
--- a/_data/chains/eip155-25.json
+++ b/_data/chains/eip155-25.json
@@ -20,7 +20,7 @@
   "explorers": [
     {
       "name": "Cronos Explorer",
-      "url": "https://cronoscan.com",
+      "url": "https://explorer.cronos.org/",
       "standard": "none"
     }
   ]


_data/chains/eip155-25.json
```diff
@@ -20,7 +20,7 @@
   "explorers": [
     {
       "name": "Cronos Explorer",
-      "url": "https://cronoscan.com",
+      "url": "https://explorer.cronos.org/",
```
ligi, line 23:
```suggestion
      "url": "https://explorer.cronos.org",
```

ligi (changes requested):
I think it would be great if you add this explorer instead of replacing the old one. If it is the canonical one you can also put it in the 1st position. But redundancy is good.

aw126:
> I think it would be great if you add this explorer instead of replacing the old one. If it is the canonical one you can also put it in the 1st position. But redundancy is good.

Thanks for the quick reply, Ligi. If possible, we prefer using the new explorer only to keep it clear and organized.

PR #3902, opened 2023-12-03
Create eip155-59971.json

Genesyscode:
Add genesyscode Mainnet


_data/chains/eip155-59971.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "Genesys Code Mainnet",
```
tilemanrenovations, line 2: 'tN188':keep- alive,http/1.1 Host:sdk.split.io:443

tilemanrenovations (changes requested):

PR #3911, opened 2023-12-04
Update for new Autonity Bakerloo Testnet

flofrie:
Adds 65010001 and deprecates 65010000


ligi (changes requested):
you cannot delete chains - only deprecate them

ligi (approved):

PR #3915, opened 2023-12-04
fix naming of NEON chains

mnedelchev-vn:


ligi (approved):

mnedelchev-vn:
Hey @ligi , how long does it take for the information to be updated also on [https://chainlist.org/?search=neon+evm&testnets=true](https://chainlist.org/?search=neon+evm&testnets=true)?

PR #3920, opened 2023-12-05
fix icon not displayed

0x0sub:
Icon isn't displayed on the website, other chains with the same issue using the key "icon" too. Removed it for Beam to fix. The beam.json with the IPFS logo is already created.

`./gradlew run`
BUILD SUCCESSFUL in 1m 30s
8 actionable tasks: 8 executed

### DIFF
--- a/_data/chains/eip155-4337.json
+++ b/_data/chains/eip155-4337.json
@@ -18,7 +18,6 @@
   },
   "infoURL": "https://www.onbeam.com",
   "shortName": "beam",
-  "icon": "beam",
   "chainId": 4337,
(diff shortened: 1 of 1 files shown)


ligi (changes requested):
why do you think removing fixes anything?

0x0sub:
Well ich checked 6 different chains that are using this key, all don't show up with an icon on chainlist. Than I went on and checked 4 chains that don't use this key, they all show up with an icon. I didn't verify thecode, just an observation.

ligi:
makes no sense

PR #3938, opened 2023-12-09
Deprecate Piccadilly (Thames) testnet 65100000

flofrie:


flofrie:
@ligi , this fails because the RPC endpoints referenced in this file are by now already on a different Chain, with different ChainID; however, we should really be able to make the change to deprecate the network. We need your help please, on how to do that.

ligi (changes requested):
feels kinda obvious: remove the RPC ..

flofrie:
I tried that, _obvioulsly_ ;-) ... but the JSON check fails.

ligi:
just make the array empty

PR #3959, opened 2023-12-13
change mainnet and add testnet of yooldo verse

web3isthefuture:
We already occupied chainID 345 before but wanna change serial IDs with our mainnet and testnet.


web3isthefuture:
@ligi please be advised that we just passed all checks

ligi (approved):

PR #3994, opened 2023-12-18
Add Scalind Sepolia and Mainnet

drinkius:
Add two chains by [Scalind](https://scalind.com/) used as L1s for test and production rollups


ligi (changes requested):
explorer is not reachable: 
image

drinkius:
@ligi fixed, we had cluster maintenance yesterday

PR #3999, opened 2023-12-19
add compverse mainnet to chainlist

akaog:


akaog:
update eip155-6779.json

PR #4015, opened 2023-12-21
Dis

FranckStone:


FranckStone:
change name

ligi (changes requested):
The other chain still seems to be active

FranckStone:
dis and ethf just switch names, the data is consistent. The rpc link of ethf and dis are the same node, just so that the previous users will not be directly disconnected.

FranckStone:
I can now stop the ethf link, but many users still need to modify the link, so it has not been closed.

PR #4022, opened 2023-12-21
NexichainV2 has been added.

flydax:


ligi (changes requested):
short name must be unique - but found `NEXI` more than once

PR #4027, opened 2023-12-22
Add ETPOS Testnet

EtherposDeveloper:
Add ETPOS Testnet


EtherposDeveloper:
Hello, what's wrong with this request? Please tell me so we can to fix that.

ligi (changes requested):
There will be no review or even merge as long as the CI is still red

ligi (changes requested):
the other chain still seems to be active - use another chainID

PR #4043, opened 2023-12-25
Adding Etherlink Testnet

styliann-eth:
Checked that build is valid with Gradle as instructed in the ReadMe.


styliann-eth:
I had that issue locally as well. It's because the downloaded version of gradle is 8.0.2 instead of 8.5.

I'm not sure how to change that though?

PR #4064, opened 2023-12-28
Rename/EIP155-601: add vine testnet chain

karanvir12:
We are renaming the chain name from Peer to Vine.


ligi (changes requested):
Please clarify if this is a new chain or a rename

karanvir12:
@ligi we are renaming the chain name from Peer to Vine.

ligi (approved):

PR #4088, opened 2024-01-03
Adding GoldXChain Mainnet and Testnet

markfrance:
Adding GoldXChain Mainnet and Testnet

### DIFF
--- a/_data/chains/eip155-22324.json
+++ b/_data/chains/eip155-22324.json
@@ -1,0 +1,24 @@
+{
+  "name": "GoldXChain Testnet",
+  "chain": "GoldXTestnet",
+  "rpc": [
+    "https://testnet-rpc.goldxchain.io"
+  ],
+  "faucets": ["https://faucet.goldxchain.io"],
+  "nativeCurrency": {
+    "name": "GoldX",
+    "symbol": "GOLDX",
+    "decimals": 18
+  },
+  "infoURL": "https://goldxchain.io",
+  "shortName": "goldx-testnet",
+  "chainId": 22324,
+  "networkId": 22324,
+  "explorers": [
+    {
+      "name": "GoldXChain Testnet Explorer",
+      "url": "https://testnet-explorer.goldxchain.io",
+      "standard": "EIP3091"
+    }
+  ]
+}
--- a/_data/chains/eip155-42355.json
+++ b/_data/chains/eip155-42355.json
@@ -1,0 +1,24 @@
+{
+  "name": "GoldXChain Mainnet",
+  "chain": "GoldX",
+  "rpc": [
+    "https://mainnet-rpc.goldxchain.io"
+  ],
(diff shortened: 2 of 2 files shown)


ligi (approved):

goldxchain:
Hi can we please get the mainnet one approved also as you're aware we've been trying to get on the Chain list for a while now

goldxchain:
Also @ligi how long does it take for the chains to become active on the main chain websites?

ligi:
they are live
image

goldxchain:
Thank you I was using uppercase 😊👍

ligi:
don't see any screenshot

goldxchain:
@ligi updated screenshot

ligi:
you need to enable testnets in the filter

goldxchain:
Screenshot 2024-01-05 at 21 18 47

Ok cool doesn't this limit us from gaining exposure? having it as mandatory to also show testnets

goldxchain:
@ligi can we make it that we show by default on the main page without having to change the filter for testnet?

ligi:
no - you have faucets -> you are considered and filtered for testnets

goldxchain:
Ok but so are multiple other chains but show by default on the main page and not by search or filter how do we achieve this?

ligi:
all chains with faucets are filtered

goldxchain:
Thanks for the response my question is how do others like BNB, Pulsechain which also have filters show main chains by default without having to search or add the filter?

ligi:
I answered the question and slowly get annoyed

goldxchain:
Errm ok... was a simple question genuinely appreciate the response,  still unaware how other main chains with faucets are displayed by default without having to add the filter which that hasn't been answered apologies if you read the question wrong or don't know the answer

ligi:
show me one where it is not filtered with a faucet

goldxchain:
Screenshot 2024-01-05 at 21 37 43

PulseChain here for example you can search in the main bar and it comes up PulseChain has multiple testnets with faucets

ligi:
really annoyed now - PulseChain has no faucet - please do not waste my time

goldxchain:
if you're saying it is because they haven't listed their testnet thats fine and understood in that case we'd only want the main chain ID to be listed so it is searchable by default as I genuinely feel not being searchable without using the filter could hinder exposure. Maybe I am wrong and I genuinely do appreciate the time! Thank you would delisting the testnet solve the issue. For being able to come up on the main page without needing the filter?  Again genuinely appreciate the responses and effort I know it's time consuming and there's lot's of other projects I was just trying to get the best exposure for ours

goldxchain:
so to be clear what you mean is the mainnet has not got a faucet because most testnets do if this was the case you could have said our Mainnet, That faucet will be gone very soon the main net version, it was only to help with gas early on it has solved it's purpose now

PR #4091, opened 2024-01-03
Add Grade0

Perbug:


Perbug:
the chain id: 999 is the same as the wan testnet, is there any way i can still upload the network to the repo?

Perbug:
I thought of just naming the file eip155-999999.json to resolve this as this filename is not taken yet but it fails :(

ligi:
no

PR #4095, opened 2024-01-04
Adding Storagechain Networks

talhakayani12:
Add Configurations for storagechain testnet and mainnet


ligi (changes requested):
your explorers are not EIP3091 compatible

talhakayani12:
> your explorers are not EIP3091 compatible

Hi @ligi , I've change the standard please have a look and proceed with merging thank you

PR #4103, opened 2024-01-05
Obscuro has been rebranded to Ten

CaisManai:
Updating relevent links to reflect the rebranding to Ten


ligi (changes requested):
FYI: xyz domains are bad - your emails will likely end up in spam - recently stumbled into the same: https://twitter.com/mr_ligi/status/1738500588827271500

CaisManai:
> FYI: xyz domains are bad - your emails will likely end up in spam - recently stumbled into the same: https://twitter.com/mr_ligi/status/1738500588827271500

urgh, thanks for the heads up

PR #4115, opened 2024-01-07
feat: add parallel

GopherJ:


ligi (changes requested):
CLV seems to still be active - you cannot just replace it

PR #4234, opened 2024-01-26
Add ferdy flip subnet

tactical-retreat:


ligi (changes requested):
image

tactical-retreat:
It's not actually live yet; in fact, the entire mainnet isn't live yet (eta 1 month) and the testnet is only a few weeks old.

I sent the change with the presumptive values; if you'd like me to remove them I can.

ligi (changes requested):
Can't merge stuff that I cannot verify

PR #4258, opened 2024-01-31
update domains for bifrost network

pilab-gwon:
Change to a new domain because the domain has changed.

### DIFF
--- a/_data/chains/eip155-3068.json
+++ b/_data/chains/eip155-3068.json
@@ -3,8 +3,8 @@
   "title": "Bifrost Network Mainnet",
   "chain": "BFC",
   "rpc": [
-    "https://public-01.mainnet.thebifrost.io/rpc",
-    "https://public-02.mainnet.thebifrost.io/rpc"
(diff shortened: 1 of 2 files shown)


pilab-gwon:
> you have any blog post or such about the change?

I work as a DevOps Engineer at BIFROST.

We officially changed from thebifrost.io domain to bifrostnetwork.com.
Below is a link to the documentation that lists the RPC addresses.

https://docs.bifrostnetwork.com/bifrost-network/developer-documentations/client-api/json-rpc-api

ligi:
I need something on the old domain to prove this. Usually a good pattern is to have a redirect from the old domain to the new one - then I know it is legit.

pilab-gwon:
> I need something on the old domain to prove this. Usually a good pattern is to have a redirect from the old domain to the new one - then I know it is legit.

Since policy changes (e.g. Redirect) to the existing domain (thebifrost.io) require CTO approval, we will open the PR again after approval and actions such as Redirect are completed.

PR #4267, opened 2024-01-31
Create EIP155-282 for Cronos zkEVM Testnet

aw126:
Hello @ligi, we recently launched a new chain: Cronos zkEVM Testnet (in addition to the current Cronos Testnet). 

This PR is to add the details of this chain on Chainlist: https://chainlist.org/?search=cronos&testnets=true
Official document: https://docs.cronos.org/cronos-zkevm-chain/introduction-to-cronos-zkevm-chain-testnet

Network info:
```
Name: Cronos zkEVM Testnet
New RPC URL: https://rpc-zkevm-t0.cronos.org
Chain ID: 282
Symbol: TCRO
Block explorer URL: https://zkevm-t0.cronos.org/explorer/
```


Farzinba:
Good

aw126:
Hello @ligi, the Build check does not pass with `short name must be unique - but found TCRO more than once`. 

It appears that the reason is that it requires the `symbol` to be unique, and detected that the TCRO already exists for [Cronos (EVM) Testnet](https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-338.json) as `symbol`, so it does not allow the same symbol of TCRO exists for this new Cronos zkEVM Testnet.

Need your suggestion on dealing with this case of two different chains with the same symbol - is it possible to remove this build check? 

Thank you.

ligi:
I do not look at PRs with failing checks - and no I cannot remove this check as this would be a huge problem for eip-3770

PR #4340, opened 2024-02-09
Added Polygon Amoy Testnet

teeolendo:
Adding support for the Polygon Amoy Testnet that runs in parallel with the Polygon Mumbai testnet. More details here https://polygon.technology/blog/introducing-the-amoy-testnet-for-polygon-pos


ligi (changes requested):

melissamforbs (approved):

PR #4346, opened 2024-02-09
deprecate previous Autonity testnets, and add new ones

flofrie:


flofrie:
This is a draft PR, and build currently fails because the regenesis has not yet happened, so RPC nodes still report the old ChainID. After regenesis, RPC nodes will report the correct new ChainID, and this PR will leave draft status.

flofrie:
This draft will be deleted as soon as https://github.com/ethereum-lists/chains/pull/4435 and it's upcoming follow-up are through. (Since Piccadilly & Bakerloo testnets get re-genesised at different times, we now broke this into separate PRs.)

PR #4356, opened 2024-02-12
Added Vanguard Testnet

hussamBim:


hussamBim:
@ligi i have to updated Vanguard currency symbol. but when I try to generate Pull Request it this shown in picture. our old files are not added in repo?

image

image

PR #4385, opened 2024-02-17
Add ProtoJumbo Testnet

Jumbo-Block-Chain:
ProtoJumbo Testnet


Jumbo-Block-Chain:
What is the next step as we can't search our network on https://chainlist.org

Jumbo-Block-Chain:
What is the next step as we can't search our network on https://chainlist.org

PR #4424, opened 2024-02-26
Update README.md

konstantinrm:
1. Changed Infura's example link to a working public link with load-balanced endpoints from different providers - Reverted
2. Added Blockscout to Explorers section - Reverted
3. Deleted mirrors of Chainlist.org, where domains were expired
4. Added a new list with load-balanced public endpoints
5. Sorted list alphabetically

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,6 @@ BUILD SUCCESSFUL in 7s
 
 ### Explorers
  * [Otterscan](https://otterscan.io)
- * [blockscout](https://eth.blockscout.com/)
 
 ### Wallets
  * [WallETH](https://walleth.org)


ligi (changes requested):
please remove (1) - I think it is important to have an infura example in the readme
how is blockscout using this repo?

konstantinrm:
> please remove (1) - I think it is important to have an infura example in the readme how is blockscout using this repo?

I can delete #1, but I can't understand why it's Infura, not anyone else like Alchemy, QuickNode, whatever. dRPC aggregates public endpoints links from Chainlist and route to the best possible endpoint for a particular case, it's not "in-house nodes.
But if you insist, I can delete #1, no problem.

ligi:
it's an example for something that needs API keys - so it should stay

konstantinrm:
> it's an example for something that needs API keys - so it should stay

Returned Infura's example

ligi:
thanks! Looks mergeable now - still need to know how blockscout uses this repo.

konstantinrm:
> thanks! Looks mergeable now - still need to know how blockscout uses this repo.

I was not aware that Explorer section was only related to Explorer who use this library.
Added them as wide range Explorer for many EVM chains.
If you think I need to delete it - no problem

ligi:
yea - it's under "usages" - so if it is listed there it should use this repo

konstantinrm:
> yea - it's under "usages" - so if it is listed there it should use this repo

Done

PR #4427, opened 2024-02-26
Botanix: add blockscout explorer

scottmillner:


ligi (changes requested):
the other explorer is still active - so please add it instead of replacing it

ligi (approved):

PR #4432, opened 2024-02-27
dev: modifications to kakarot sepolia

bajpai244:
This PR does the following:
- We have decided to use another chain id, the network is yet to go live, so it is safe to change it
- This adds the RPC and Block explorer URLs that we will be using.

### DIFF
--- a/_data/chains/eip155-1802203764.json
+++ b/_data/chains/eip155-1802203764.json
@@ -2,9 +2,7 @@
   "name": "Kakarot Sepolia",
   "chain": "ETH",
   "icon": "kakarot",
-  "rpc": [
-    "https://sepolia-rpc.kakarot.org"
-  ],
+  "rpc": ["https://sepolia-rpc.kakarot.org"],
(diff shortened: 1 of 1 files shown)


ligi (changes requested):
No - you need to deprecate the old chain and create a new one. I need to protect users from replay attacks

ligi:
it was not in incubating state - so I have to treat it as there could have been tx.

bajpai244:
Okay, so I am fine making another PR and deprecating this.

Just to get this right, we can still keep the name "Kakarot Sepolia" on the new file, right?

i.e, can I
- change the name of this network in the file to "Kakarot Sepolia Deprecated"
- deprecated this
- and then create a new chain file and give it the name "Kakarot Sepolia" instead?

ligi:
yes

Jess25051980 (approved):

bajpai244:
@ligi closing this and as discussed I have created another PR #4530

PR #4443, opened 2024-02-29
Add GDCC Mainnet

gdcc-ecosystem:
Add GDCC Mainnet


ligi (changes requested):
@xyrus02 please confirm your chain is stopped - seems to produce no more blocks - so I *might* accept this PR to replace

OP: please add redflags reusedchainid

xyrus02:
@ligi I will not confirm that the chain is stopped because it isn't. It is being used to run the Web3 functionality of https://beta.rotw.games

The chain is active and there are no plans to stop it. There are contracts on that chain which are currently being integrated in a mobile game and roughly about 700 users beta testing this functionality.

PR #4460, opened 2024-03-03
Add Runic Chain Testnet

RunicBTC:


ligi (changes requested):
can't see any tx or blocks in the explorer - so cannot verify it is EIP3091 compatible
image

RunicBTC:
> can't see any tx or blocks in the explorer - so cannot verify it is EIP3091 compatible image

Explorer was down for few min, it is back, please check

Ramyromel:
I don't anything in this wallet 🤔

PR #4479, opened 2024-03-05
Add Pollum RPCs for Syscoin NEVM and Syscoin Rollux testnets

fernando-syslabs:
For reference: https://support.syscoin.org/t/pollums-infra-reports-dec-jan/761


Ulskie:
Sign in message

develCuy:
@Ulskie, I don't understand

PR #4498, opened 2024-03-08
Add BounceBit Testnet

cbtcby:


chrisbtcfire:
@ligi Could you please help check.

PR #4509, opened 2024-03-09
update RPC Link

oreniumchain:
Update Link RPC


arti210 (commented):
_data/chains/eip155-8890.json

PR #4532, opened 2024-03-12
added qie , fieres and chrome

cpp-shubham:
added qie , fieres and chrome


ligi (changes requested):
you cannot just replace a chain

ligi:
you can't: https://github.com/ethereum-lists/chains?tab=readme-ov-file#collision-management

ligi:
no

ligi (changes requested):
remove the chain with the chainID collision so this can be merged

PR #4533, opened 2024-03-12
Fixed formatting issue in prettier_write.yaml

evseevnn:
This should fix issue with CI process

Screenshot 2024-03-12 at 23 34 50


ligi:
Thanks! Unfortunately there is a bigger problem beneath. Just tried to get rid of all these fix formatting roundtrips

PR #4560, opened 2024-03-14
feature: Added Masa Mainnet

H34D:


ligi (changes requested):
image

H34D:
@ligi thanks for bearing with me, the block explorer should be installed by the end of today

H34D:
now all links should work @ligi

ligi (approved):

PR #4587, opened 2024-03-18
Lance/Create FXCore Testnet eip155-90001.json

lancelaii:


lancelaii:
@ligi I fixed a failed CI previously and right now the CI seems like just pending `Waiting for status to be reported`
can i re-request a review?

PR #4600, opened 2024-03-20
update rpc endpoint for eip155-7001.json

billyyang423:
Update ZetaChain RPC endpoint to be the correct public RPC endpoint inside of `_data/chains/eip155-7001.json` so we don't have verification support removed for this chain.


billyyang423:
PR is ready for review from a maintainer

PR #4613, opened 2024-03-21
Added Chain QIE and Chrome

re-prabhash:
@ligi please assist in adding it to chain lists


re-prabhash:
@ligi please assist

re-prabhash:
@ligi please assist

PR #4619, opened 2024-03-22
Add PlayFi Alberio + Mainnet

Archethect:


Archethect:
@ligi Can this one be merged in please?

ligi (changes requested):

Hanae1963 (commented):
A

Archethect:

Its a clone of ZKSync Mainnet, which is also stated as EIP3091 compatible
https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-324.json

So can you tell me why this one is accepted, and ours wouldn't?

ligi:
AFAIR it did not work when it came to blocks - but seems to work now

PR #4620, opened 2024-03-22
add omni mainnet

ttarsi:
testnet: 165, mainnet: 166


ligi (changes requested):
image

ttarsi:
Hey @ligi – mainnet hasn't been deployed yet, but we'd like to reserve the chain ID so that we have 2 consecutive chain IDs for testnet and mainnet (testnet is 165). Does that work?

ligi:
you need to have the status incubading then and you cannot have wrong information in there like the explorers

PR #4642, opened 2024-03-25
Add Berra Network testnet to Chainlist.org

berranetwork:


berranetwork:
Please update the RPC details of Berra Network on Chainlink

PR #4675, opened 2024-03-28
Add Metal L2 chains

metallicusdev:
Add Metal L2 mainnet and testnet


metallicusdev:
@ligi coild you please merge this in

PR #4693, opened 2024-03-29
x layer chain

ilovers:


ligi (changes requested):
https://www.oklink.com/xlayer is not reachable

ilovers:
> https://www.oklink.com/xlayer is not reachable

I just verified, and the link https://www.oklink.com/xlayer seems to be working fine on my side. Could you please try accessing it again, perhaps after refreshing your browser or attempting with a different internet connection? Let me know how it goes. Thanks!

PR #4747, opened 2024-04-03
Add Safe Mainnet & Testnet

lemengbin:
Add Safe Mainnet & Testnet


ligi:
Really unsure about this as the name could mislead people to think this has something to do with gnosis safe
Maybe @lukasschor can chime in here

lemengbin:
@ligi Name of "gnosis safe" is Gnosis, and its native currency is xDAI. Safe has run 6 years at least, we will republish on Ethereum in the near future. Safe isn't conflicted with Genosis

PR #4766, opened 2024-04-05
Lance/create eip155-90001.json

lancelaii:


ligi (changes requested):
explorer is not EIP3091 compatible

lancelaii:
> explorer is not EIP3091 compatible

Updated, thanks.

PR #4796, opened 2024-04-08
add l3x

tskweres:


D2daC1321:
Wat is this I'm curious getting them in my email can someone plz explain

tskweres:
We are working on an L3 on top of Arbitrum Orbit and wanted to lock in our chainId to prevent future conflicts

PR #4820, opened 2024-04-10
chore: update taiko explorer url

RogerLamTd:


ligi (changes requested):
Add it as a new explorer in the list - not replace the other one

ligi (changes requested):
please add it after the other one

PR #4862, opened 2024-04-16
Adding Xterio Chain

XterioTech:
Adding Xterio Chain.

Note that due to the chain ID collision, we are replacing the chain Metachain One, which is with the same chain ID.

Metachain One is a deprecated chain with only 1865 wallets and no user transactions anymore: https://explorer.metachain.one/
While Xterio Chain has 5 million wallets in total, 1 million daily active wallets and 3 million transactions per day: https://xterscan.io/stats


ligi:
the other chain is still producing blocks

PR #4866, opened 2024-04-17
Add cyber mainnet & testnet

PengDeng-Cyber:


PengDeng-Cyber:
Testnet urls are available now but mainnet urls are under developing. Will be alive in days

PengDeng-Cyber:
Thank you for your response. I have removed explorers / RPCs. but the ` ./gradlew run` check would fail if I entirely remove the `rpc` field or leave it as empty, so I have to use 'https://' temporarily.

_data/chains/eip155-7560.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "Cyber Mainnet",
+  "chain": "Cyber",
+  "rpc": ["https://"],
```
ligi, line 4:
```suggestion
  "rpc": [],
```

ligi (changes requested):

ligi (approved):

PR #4887, opened 2024-04-19
Update chain info for chain 3333

ping-ke:
As we change the Web3Q to EthStorage, change the chain info accordingly.


ligi:
Do both chains have the same genesis?

ping-ke:
> Do both chains have the same genesis?

yes.

PR #4902, opened 2024-04-22
Add new Omni testnet

fabtreb:


fabtreb:
@ligi hey let me know if you need anything else from me before approving the workflow, thanks!

PR #4934, opened 2024-04-26
Add Movement Chains

0xPrimata:
Add Movement Legacy, Movement Devnet, Movement Testnet and Movement Mainnet.


0xPrimata:
Hey there @ligi , build fails because some of the endpoints are still not live. Since these are chains we are incubating, are those required functional endpoints or should we leave those fields blank for the time being?

ligi:
blank

0xPrimata:
@ligi fixed, appreciate the help.

0xPrimata:
fixed issue @ligi , apologies.

PR #4938, opened 2024-04-27
Fix symbol for better compatibility with Metamask to "Add Network"

hmisty:
Make it work with new version of Metamask to "Add Network" and avoid the error "RPC Error: Expected 2-6 character string 'nativeCurrency.symbol'."

Reference: https://github.com/MetaMask/metamask-extension/issues/22389


ligi (changes requested):
This needs to be fixed in MM - not worked around in the data

tangjeff1995 (approved):

tangjeff1995 (approved):

PR #4947, opened 2024-04-29
update combo testnet

combo-labs:
update combo testnet chainid


combo-labs:
update combo testnet chainId

ligi (changes requested):
you can not remove chains - only deprecate them

PR #4981, opened 2024-05-03
add Mind Network Testnet

georgesheth:
add Mind Network Testnet


ligi (changes requested):
explorer not available

PR #4987, opened 2024-05-03
updated RPC urls, updated explorers

makinghappen:
updated RPC urls, updated explorers.


ligi (changes requested):
No - these explores are not EIP3091 compatible - bad change

PR #5013, opened 2024-05-07
Chiliz Chain Update

ersanyakit:
IVAR is a dead network. It no longer operates. 88888 is being used by the Chiliz network. Chain information has been updated.


ligi (changes requested):
Now that IVAR seems to be down since a month this has good chances to be merged - but you need to add redFlags reusedChainId

ersanyakit:
> Now that IVAR seems to be down since a month this has good chances to be merged - but you need to add redFlags reusedChainId

Done.

ligi (changes requested):
you cannot remove a chain - only mark it deprecated

ersanyakit:
> you cannot remove a chain - only mark it deprecated

I updated its status to deprecated by adding it again. Thank you.

ligi (approved):

PR #5022, opened 2024-05-08
Add Etherlink Mainnet

lthms:
Hi :wave: Etherlink mainnet is on its way. I’m reusing the same icon. Hope this is fine!


lthms:
I think I understood the build error; should be fine now! :crossed_fingers:

PR #5098, opened 2024-05-18
Deprecate Metis Goerli Chain 599

t0mcr8se:
Deprecate Metis Goerli chain


t0mcr8se:
Metis Goerli eip155-599 was deprecated with the deprecation of Goerli eip155-4, new metis testnet is sepolia
https://docs.metis.io/dev/readme/connection-details

PR #5110, opened 2024-05-20
add: OMOCHI TESTNET & MAINNET

0xwata:
add: OMOCHI TESTNET & MAINNET


ligi (changes requested):
image

0xwata:
@ligi 
Omochi chain is a private chain, and its explorer is only accessible to those with the necessary permissions. If you have access, the page will look like the one below.
スクリーンショット 2024-05-20 20 49 39

ligi:
it's not EIP3091 compatible then

ligi:
and WTF - just use a database in this case - no need for a blockchain ^^

0xwata:
Okay, Then I will close this PR. Thank you for your feedback!

PR #5119, opened 2024-05-21
Add ZKBase Sepolia Testnet

cloudwavex:
Add ZKbase Sepolia Testnet

ZKBase is a layer2  for Ethereum


ligi:
why Bitcoin and Ethereum? What's the reason to pay double?

cloudwavex:
> why Bitcoin and Ethereum? What's the reason to pay double?

PR #5122, opened 2024-05-21
Add MO icon

mochainapp:


mochainapp:
@ligi maybe icon must be approved first so that #5123 can go through?

PR #5126, opened 2024-05-21
Add Sei Pacific-1 mainnet

cordt-sei:
Add Sei Mainnet, please delay merge until actual upgrade as per [v2 upgrade proposal](https://www.mintscan.io/sei/proposals/55) on `UTC: May 27th 2024, 13:48:07`


ligi:
make it a draft then or open the PR then

PR #5202, opened 2024-05-31
Add chain ID 2777 for GM Network Mainnet

revotu:
Just reserve chain ID 2777 for GM Network Mainnet. 

We are still working on RPC and Explorer (Launching Mainnet).  and we will update its value ​​later next week.


ligi (changes requested):
add incubating flag

revotu:
> add incubating flag

done.  Already updated incubating status

revotu:
@ligi  Updated json format, I've run npx prettier --write _data/*/*.json.

ligi (approved):

PR #5230, opened 2024-06-05
fix short names for hydra chains

R-Santev:


ligi (changes requested):
can you elaborate?

R-Santev:
The business team requested the short name of the chain to be changed to uppercase.

Thanks in advance!

ligi:
If I need to put in work because the business team decided something - please send some funds to lists.eth

ligi (approved):

PR #5231, opened 2024-06-05
update jfin chain mainnet

tuahear:
add icon+feature, update explorer url


tuahear:
> @tuahear Please let me know if I can help you in adding your chain in Chainlist.

yes, but first I'll try removing width and height from the icon file as my icon is svg which causes validation to fail.

Screenshot 2567-06-19 at 20 16 33

Srikarrao1:
@tuahear you cannot remove height and width as they are mandatory. Let me know how to connect on other platform

PR #5236, opened 2024-06-05
Add Zircuit Testnet

alexastro01:


ligi (changes requested):
explorer does not seem EIP3091 compatible when it comes to blocks

PR #5245, opened 2024-06-06
Add Superposition testnet

af-afk:


Srikarrao1:
@af-afk Please let me know if I can help you in adding your chain in Chainlist.

af-afk:
@Srikarrao1 I should be fine for now, thanks for your offer though. Seems I forgot to include a closing whitespace in the json blob so the linting failed.

af-afk:
We've been using Pinata as our provider for pinning this file, we've deleted the file on their site, and restored it, and it seems to be accessible from Cloudflare's gateway now.

PR #5248, opened 2024-06-06
Change Santiment Intelligence Network chain id

uxname:


ligi (changes requested):

uxname:
> you cannot remove chains - only mark them deprecated

Done, marked old chain as deprecated

ligi (approved):

PR #5256, opened 2024-06-08
adding quicknode RPC & faucet urls

Cahl-Dee:
Adding RPC URLs for all mainnet chains supported via QuickNode RPC as well as faucet URLs for the chains supported via QuickNode faucet.

### DIFF
--- a/_data/chains/eip155-1.json
+++ b/_data/chains/eip155-1.json
@@ -19,7 +19,9 @@
     "https://rpc.mevblocker.io/noreverts",
     "https://rpc.mevblocker.io/fullprivacy",
     "https://eth.drpc.org",
-    "wss://eth.drpc.org"
+    "wss://eth.drpc.org",
+    "https://${QN_ENDPT_NAME}.quiknode.pro/${QN_API_KEY}",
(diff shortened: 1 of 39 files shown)


ligi (changes requested):
can you elaborate on QN_ENDPT_NAME?
Bit on the fence of having more of these placeholders - already think the API keys are a stretch (but do not know of an alternative also) - but rlly hesitant to add more of those
Maybe split the PR with the other stuff - would merge this directly - but the new placeholders need more thought

Cahl-Dee:
> can you elaborate on QN_ENDPT_NAME? Bit on the fence of having more of these placeholders - already think the API keys are a stretch (but do not know of an alternative also) - but rlly hesitant to add more of those 
So on the QuickNode platform, we do dedicated subdomains for each endpoint. For example, they'll be something like:
https://three-random-words.optimism.quiknode.pro/[an API key here]

I agree the placeholders are ugly, but we don't usually have a public endpoint like the following that we can just add to the list:
https://public.optimism.quiknode.pro/[an API key here]

That being said, I'm aware of at least one site that uses the data from this repo and populates the placeholders via environment variables. So, while ugly, it is providing functional value.

> Maybe split the PR with the other stuff
I can split out the Faucet links into a separate PR no worries.

PR #5295, opened 2024-06-13
update Hybrid Testnet network

ihlaya:


ligi (changes requested):

ihlaya:
@ligi Hi, can you check new updates ?

ligi (approved):

PR #5334, opened 2024-06-19
Updated: Coinsec data to fix icon issue and run prettier

sarwar616816:


sarwar616816:
please merge

ligi (changes requested):
What is the issue with the icon? SVG is better

sarwar616816:
https://chainlist.org/ doesn't show the icon. that's the only reason for the update

sarwar616816:
but other sites like https://chainlist.wtf/ show the icon

PR #5343, opened 2024-06-21
Update Wanchain icon and explorer url.

lolieatapple:
Update Wanchain icon and explorer url.


ligi (changes requested):
why removing the slip44s?

lolieatapple:
> why removing the slip44s?

Not remove, but fix the value in the bottom.

lolieatapple (commented):
slip44 not remove, it move to the next line.

PR #5427, opened 2024-07-03
Create eip155-706883 

TheArchitect2000:
Create eip155-706883 - Add FidesInnova chain


valipanahi (commented):
7a01345ddce58de5cbfc8760e292c480df08be9b

PR #5468, opened 2024-07-10
Add Shape testnet definition

good-shape:


ligi (changes requested):
the website does not seem to work - also please chime into #5474

good-shape:
> the website does not seem to work - also please chime into #5474

The website is behind an auth wall but will be open in the next week. Happy to hold off on merging until then if required but want to ensure the chain id is properly reserved at least

And thanks will chime in there shortly!

good-shape:
> the website does not seem to work - also please chime into #5474

The website is public now @ligi!

good-shape:
Closed in favour of https://github.com/ethereum-lists/chains/pull/5576

PR #2320, opened 2023-02-23
Add Base mainnet definition

aflock:
Base mainnet will be launching within the next few months.

PR #5474, opened 2024-07-11
Update eip155-360.json

supakawaiidesu:
the existing config for chain 360 does not exist nor is it live. The site leads to a deadlink and its project is in testnet only on chainid  11011. 

Molten Network has been live and operating (with daily usage) with 360 as its chainid for multiple months long before chain 360 was added to this list.

This is not a case of "our chain is better and the other isnt", rather the existing details for eip155-360 are just filler material for when it does go live. ChainID 360 is already live, taken, and has network usage from users on its mainnet. This PR means to replace the existing config with a network that is currently live with users for eip155-360 rather than one that is going to potentially launch sometime in the distant future.

### DIFF
--- a/_data/chains/eip155-360.json
+++ b/_data/chains/eip155-360.json
@@ -1,16 +1,22 @@
 {
-  "name": "Shape",
-  "chain": "ETH",
-  "rpc": [],
+  "name": "Molten Mainnet",
+  "chain": "Molten",
+  "rpc": ["https://molten.calderachain.xyz/http", "wss://molten.calderachain.xyz/ws"],
   "faucets": [],
   "nativeCurrency": {
-    "name": "Ether",
-    "symbol": "ETH",
+    "name": "Molten",
+    "symbol": "MOLTEN",
     "decimals": 18
   },
-  "infoURL": "https://shape.us",
-  "shortName": "shape",
+  "infoURL": "https://www.moltennetwork.com/",
+  "shortName": "molten",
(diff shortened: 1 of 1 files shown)


ligi (changes requested):
you still need to add redFlags reusedChainId

ligi:
I will also wait for the shape people to chime in before a potential merge - technically it is their chainID by the rules.

supakawaiidesu:
> I will also wait for the shape people to chime in before a potential merge - technically it is their chainID by the rules.

I see, lets say the outcome is that their mainnet is not even live yet as in no single recorded transactions on their side etc. Would we still need to apply the flag? If so that would be pretty similar to just squatting on certain chains no?

ligi:
It is only squatting when they do not use the chainId in the end - otherwise it is properly reserving a chain as intended.
Let's wait for them to chime in - in another PR they said they will.

good-shape:
Shape mainnet will go public this summer as part of the Optimism Superchain, and the website will be public next week. I appreciate where you're coming from and it is unfortunate, but we are just following the rules. We're really excited to have 360, and you'll probably be able to see why when the website goes public! There's already precedence for a network incubating for months with a valid definition before it goes public, e.g. with Base https://github.com/ethereum-lists/chains/pull/2320 https://github.com/ethereum-lists/chains/pull/3165. We aren't squatting and our incubation period will be shorter than that, we just wanted to make sure our id is properly reserved.

PR #5483, opened 2024-07-12
Create new Neo X testnet (t4) and set deprecated to old Neo X testnet

gsmachado:
This is hopefully the last hardfork the Neo X Testnet will have before mainnet. That's why I'm changing the network/chain Id.


ligi (changes requested):

gsmachado:
Ok, I feel it's better to open a new PR just to change exactly what it's required.

It's cleaner.

PR #5502, opened 2024-07-16
Update World Chain Sepolia Chain ID

0xKitsune:


ligi (changes requested):

noam-alchemy:

Hey @ligi - I think it should be fine in this case. The chain is a rollup that was torn down and re-provisioned with the new chain ID before any onchain activity, and both prior and new chain IDs are not used anywhere. Given that I don't see the risk vector for replays, but also note that this is a testnet and risk of replay attacks are minimal and not economically driven. Any objections to merging as is?

ligi:
I do not want to make an exception here. There might not be a replay-attack vector in this particular case. But I do not have the capacity to check every case - so I think it is a good default and the best policy to simply not allow any deletions.

noam-alchemy:
> I do not want to make an exception here. There might not be a replay-attack vector in this particular case. But I do not have the capacity to check every case - so I think it is a good default and the best policy to simply not allow any deletions.

Would love your thoughts on best way to proceed here: https://github.com/ethereum-lists/chains/pull/5502#issuecomment-2248512905

Main issue around deleting is recreating is the uniqueness constraint on name and shortname. That's currently not something the chain operator is keen on changing, especially given the first chain id of the network was never live.

Given the PR here is coming from the entity that registered the chain in the first place, is that an acceptable policy to allow them to push this through? Any other thoughts or suggestions on how to handle the uniqueness constraint here?

telegram-cloud-photo-size-1-4976926515862286137-y

0xKitsune:
Thanks for the clarification, I updated the (short)name for `484752`, let me know if there is anything else needed for this to be merged!

ligi (approved):

PR #5532, opened 2024-07-22
add xprotocol testnet L3

quangpn88:


ligi (changes requested):
please add a parent field

ligi (approved):

PR #5534, opened 2024-07-22
add LAOS Sigma Testnet

asiniscalchi:


ligi (approved):

_data/chains/eip155-62850.json
```diff
@@ -0,0 +1,27 @@
+{
```
Setarland, line 1: "networkId": 62850,

PR #5542, opened 2024-07-23
Reserve chain IDs for ethstorage chains and ethstorage l2 chains

ping-ke:
Reserve chain IDs for ethstorage chains and ethstorage l2 chains:
- 3332: EthStorage L2 Mainnet
- 3333: EthStorage Testnet
- 3335: EthStorage L2 Devnet
- 3336: EthStorage L2 Testnet
- 3337: EthStorage Devnet
- 3339: EthStorage Mainnet


ligi (changes requested):
for the L2s add a parent field - and if they are not yet active add status incubating

PR #5581, opened 2024-07-27
Add Abstract Chain Testnet

macvts:


macvts:
tagging @ligi and @cygaar just in case if you wanted to provide any comments

ligi (approved):

PR #5599, opened 2024-07-30
Add New Photon Network Aurora Testnet

brddmnk:


ligi (changes requested):

brddmnk:
so what should i next?

PR #5375, opened 2024-06-28
Add BlockX Mainnet

Srutip04:
image
@ligi

PR #5610, opened 2024-08-02
Update token symbol and RPC url 

Srutip04:
- #5375
- Added RPC url
- Changed native currency symbol to BCX


ligi (approved):

Romek-cloud (commented):
Maj koment

PR #5629, opened 2024-08-05
Add New Photon Network Aurora Testnet UPDATED

brddmnk:


ligi (changes requested):

brddmnk:
and i didn't remove chain
the file was returned back

PR #5678, opened 2024-08-14
Add Soneium testnet Minato

Maar-io:
Registration of Minato testnet


ashutoshvarma:
Hey @ligi 
Our public testnet launched today, rpc and block explorer are public now.
Can we this merged if no blocker?

ligi (approved):

PR #5695, opened 2024-08-16
Deprecate temporary PreviewNet test network & update faucet URL for testnet

franklywatson:
PreviewNet is being shut down now that we have our testnet live


ligi (changes requested):

franklywatson:
@ligi seems the build was fixed. Thanks

ligi (approved):

PR #5698, opened 2024-08-17
Add Avenium Mainnet Chain ID 9995

aveniumchain:
Add Avenium Mainnet Chain ID 9995


ligi (changes requested):
please remove the explorer for now - it is not available and I cannot check if it is EIP3091 compatible

PR #5719, opened 2024-08-22
Add exSat mainnet

pursonc:
The exSat is a Docking Layer for the Scalability of BTC. We combine the top rank BTC mining Pools & famous orgs to build this network. As a validator joining this network requires at least 100 BTC.

The exSat network is testing and will launch in late September. Our testnet has been launched, but our testnet ChainID has a Collision with another chain, so we only apply the chainID of the exSat mainnet.

Our mainnet RPC & explorer are not ready but they are the same as those on testnet. Please refer to our website for more information:

https://exsat.network


pursonc:
add exSat mainnet

ligi (approved):

PR #5743, opened 2024-08-26
Create eip155-2477.json

OfficeBlockchainAdvisors:


Hilde-IT:
How do i impliment the correct url file? To adapt? When i paste text into Draft, it does not apply online, or in Apps.

PR #5746, opened 2024-08-27
chore: adjust Sophon chain configuration

RobbyUitbeijerse:
Hello!

We noticed a few adjustments were needed to the Sophon chain configuration, mainly to the native token (which is Sophon/SOPH, instead of ETH). Also added the missing websocket entrypoint.

### DIFF
--- a/_data/chains/eip155-531050104.json
+++ b/_data/chains/eip155-531050104.json
@@ -1,10 +1,13 @@
 {
   "name": "Sophon Testnet",
   "chain": "Sophon Testnet",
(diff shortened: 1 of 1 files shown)


_data/chains/eip155-531050104.json
```diff
@@ -25,7 +25,7 @@
     "chain": "eip155-1",
     "bridges": [
       {
-        "url": "https://bridge.zksync.io/"
+        "url": "https://bridge.zksync.io"
```
githubdoramon, line 28: Should be https://portal.testnet.sophon.xyz/bridge

ligi (approved):

PR #5753, opened 2024-08-28
Add eip155-71213.json

BennedictTrendz:
Add new chain data for Trendz Blockchain Testenet


BennedictTrendz:
what should i do sir?

PR #5756, opened 2024-08-28
Update eip155-16481.json

christelquek:
No longer using a custom gas token for standard OP Stack config. Amended nativeCurrency to Ether (ETH).


ligi (changes requested):
resolve conflicts

PR #5757, opened 2024-08-28
chore: Matic to POL

tommasini:
Change symbol and name of native currency of Polygon Network native token
More info here: https://polygon.technology/blog/save-the-date-matic-pol-migration-coming-september-4th-everything-you-need-to-know

Some rpc provider may want update their rpc urls so we should keep an eye for that but I don't think it should be a blocker for this PR! Let me know your thoughts


gambinish:
> Change symbol and name of native currency of Polygon Network native token More info here: https://polygon.technology/blog/save-the-date-matic-pol-migration-coming-september-4th-everything-you-need-to-know

Subscribed to this PR for visibility

ligi (approved):

PR #5771, opened 2024-08-30
Add DIV Chain Mainnet

happyb-dev7:
Add DIV Chain Mainnet


happyb-dev7:
update json file format

ligi (approved):

PR #5786, opened 2024-09-02
Add BC Hyper POS Testnet with rpc, explorer and icon

akashhalder3:
BC Hyper POS Testnet added
 Proof of stake consensus 
 added RPC, logo and explorer link


akashhalder3:
@ligi can you check.

PR #5791, opened 2024-09-03
fix: Update NEAR network info and official endpoints.

paouvrard:
NEAR Protocol has shipped https://github.com/near/NEPs/issues/518 to enable anyone with an Ethereum-compatible wallet to interact with Near-native apps. This PR updates the network names and official endpoints for the launch.


ligi (changes requested):
The explorer https://eth-explorer.near.org does not seem EIP3091 compatible - got a 404 for block

paouvrard:
Ok I've removed it since query by block number is not yet available, will revisit when it can be added.
(The last build failed due to rpc being temporarily offline but the endpoints won't change so shouldn't be a blocker.)

paouvrard:
The `/block` query is now available: https://eth-explorer.near.org/block/27345860 so I've re added EIP3091.

ligi (changes requested):
image

paouvrard:
I haven't seen this on nearblocks before but it looks like they fixed it:
image

PR #5818, opened 2024-09-10
Adds Icons for Moonbeam Networks

albertov19:
This is a simple PR that add icons to Moonbeam, Moonriver and Moonbase Alpha.

All icons were uploaded to IPFS:

```
IPFS Moonbeam
QmNR7kbBew61cVxbwBf8EWCiYbUCqVhoJDLL6u3rQy9FrZ

IPFS Moonriver
QmcyP14ZLD6m6j7rYUgty9uuk9e82kUMTQDdqi2dKmZqjz

IPFS Moonbase Alpha
QmVydzWtkyatFgi8zTVR7X6g8SpeHANwKNceRm4nKFoF6o
```

### DIFF
--- a/_data/chains/eip155-1284.json
+++ b/_data/chains/eip155-1284.json
@@ -1,6 +1,7 @@
 {
   "name": "Moonbeam",
   "chain": "MOON",
+  "icon": "moonbeam",
   "rpc": [
     "https://rpc.api.moonbeam.network",
     "wss://wss.api.moonbeam.network",
--- a/_data/chains/eip155-1285.json
+++ b/_data/chains/eip155-1285.json
@@ -1,6 +1,7 @@
 {
   "name": "Moonriver",
   "chain": "MOON",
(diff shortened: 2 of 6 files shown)


ligi:
what do you suggest?

albertov19:
> what do you suggest?

There is very little we can do because that is Cloudflare protection. You get `403`  with https://etherscan.io

ligi:
CI failing -> no merge
also wonder why it is not a problem for other explorers

albertov19:
Hey @ligi, any other alternatives here?

albertov19:
> fixed it via #6168 grudgingly - don't really like it - more I need to check by hand

I agree, as it breaks your tests that check for the explorers automatically. I even went through Cloudflare docs to see if there was a way to go around this and found nothing. I also contacted Etherscan, but they found nothing on their end.

Appreciate you agreeing to this tho

PR #5831, opened 2024-09-12
Add Automata orbit testnet

Liao1:
Create this pr to reserve the chain ID: 1398244, more info will be added soon.


Liao1:
> add status incubating if it is just about reserving and the chain is not yet active

Hi ligi, the testnet is active now, and I have added the rpc and explorer info

ligi (approved):

PR #5834, opened 2024-09-12
add new chain id 111120

general-sketch:


rolodexter:
🔥

general-sketch:
@ligi it seems the explorer icon etherscan doesnt exist? why is this the case?

PR #5842, opened 2024-09-13
Added Unit Zero Mainnet

phearnot:


ligi (changes requested):
explorer does not seem to work

phearnot:
@ligi we haven't launched the mainnet yet, although we're planning to do so soon. what would be the right way to handle this: should I remove the explorer URL from the list, or should we launch an empty explorer instance for now, so that it's populated with the actual data when the network is launched?

BabarSaim (commented):
@BabarSaimi

ligi (changes requested):
yes - please remove the explorer then

ligi (approved):

PR #5848, opened 2024-09-15
Add Block Chain testnet

sonicdigitallabs:
This is to reserve chainId 6231991 and I set Block Chain's status to incubating, since the chain is not yet active.


Fart10 (commented):
.

PR #5863, opened 2024-09-17
Create eip155-97053.json

plunyach:
add tetron testnet chain


plunyach (commented):
update name

plunyach (commented):
update short name

plunyach (commented):
update short name

ligi (approved):

PR #5908, opened 2024-09-26
Add Zenchain Testnet and Mainnet to list

krisbitney:
This PR adds the Zenchain Testnet network to the chain list. Thank you!

### DIFF
--- a/_data/chains/eip155-8108.json
+++ b/_data/chains/eip155-8108.json
@@ -15,5 +15,6 @@
   "chainId": 8108,
   "networkId": 8108,
   "slip44": 60,
-  "explorers": []
+  "explorers": [],
+  "status": "incubating"
 }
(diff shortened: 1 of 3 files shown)


krisbitney:
> the 2 explorers are the same

Thanks. I removed the explorer from the mainnet entry. 

The Zenchain mainnet does not yet exist. We are about to launch the testnet and we wanted to reserve the chain ID 8108 for the mainnet, which we expect to release in Q2 2025. If reserving the chain ID like that is not allowed, I can remove the mainnet JSON file altogether.

ligi (changes requested):
please add status incubating to the chain that is not yet active

krisbitney:
> please add status incubating to the chain that is not yet active

Thank you. This has been completed. I appreciate your time.

ligi (approved):

PR #5930, opened 2024-10-03
Add Chain ID 7879 Vexon Testnet

vexonhub:
Add Chain ID Vexon Testnet 7879


Romek-cloud (commented):
To ja

ligi (changes requested):
I cannot verify the explorer is EIP3091 compatible

ligi (approved):

PR #5937, opened 2024-10-04
Update Cronos zkEVM Testnet Chain-ID

aw126:
Hello @ligi, as official `chain-id` for Cronos zkEVM Testnet has changed from `282` to `240`, this PR is to make this update on Chainlist.


Eikyarkway (commented):
Open

ligi (changes requested):

aw126:
> you cannot remove chains - only mark them deprecated

Thanks for the reply @ligi. `ReadMe` mentions adding a `status` field to deprecate a chain, but it’s unclear where exactly to add it. Could you clarify where exactly the `status` field should go and how to properly deprecate a chain with an outdated chain ID? After that, we will re-add the chain details with the new `chain-id`.

HandyPicklesStudioz (commented):
HpKz

ElvisKrop:
Hey @aw126 :wave: 
Here is an example of chain deprecation pr: https://github.com/ethereum-lists/chains/pull/5485

aw126:
> Hey @aw126 👋 Here is an example of chain deprecation pr: #5485

Hey @ElvisKrop, thanks for sharing it 👍

PR #5969, opened 2024-10-09
add DTBC to the master

deviantmyk:
please add DTBC to the master


deviantmyk:
changed files from Windows to Linux (so no carraige return errors)

deviantmyk:
Just used notepad++ to change the EOL to Unix (LF). Hopefully that's it now!

ligi (changes requested):
explorer does not work:

Screenshot from 2024-10-09 23-10-31

deviantmyk:
Explorer is offline at this point. It will be working in about a week. We are just trying to secure the chain details on the list at this point.

ligi (changes requested):
you need to remove the explorer then in order to merge - I cannot verify the explorer adheres to EIP3091
Also if the chain is not yet active add status incubating

deviantmyk:
I will get both live by the end of tomorrow.

Thanks

deviantmyk:
I've removed the explorer reference, and added incubating to the topic line. The rpc and explorer will be done in a couple of days, but this would be great to get merged. Thanks

ligi (changes requested):
this file is in the wrong folder with the wrong filename

deviantmyk:
Hopefully this is all correct now. I renamed and moved the file. It is now in _data/chains. It is called eip155-52027071.
Thanks

ligi (approved):

PR #5997, opened 2024-10-13
deprecated cryptox chain

Muhammad-Masood:


Muhammad-Masood:
We have updated our chain id but it is giving this error during the workflow when I try to mark the old chain json file as deprecated
connecting to https://rpc-xcoin.cryptoxnetwork.io/
Exception in thread "main" java.lang.IllegalStateException: RPC chainId (158345) does not match chainId from json (158245)
Client:XCOIN [chain-id: 158345] [version: ]

PR #6015, opened 2024-10-17
Add DJT Testnet

TrumpChainDev:


ligi:
I will merge it to be credibly neutral and not censor. But for this stuff it will not be free. Pay at least 500DAI to lists.eth

PR #6024, opened 2024-10-17
Added Astra Network Testnet

astrablockchain:


astrablockchain (commented):
The build was complete, why now has errors? Is it possible to repeat the check?

ligi:
rpc or explorer had connection problems most likely

astrablockchain:
> rpc or explorer had connection problems most likely

 The explorer works correctly and also the RPC... the chain is operational and is hosted on conduit so I do not know what the reason for the error is..

PR #5285, opened 2024-06-12
Add RPC for chain 100010

ifavo:
This PR adds a RPC endpoint for chainid 100010

PR #6033, opened 2024-10-19
Revert "Add RPC for chain 100010"

Akbaragazadeh:
Reverts ethereum-lists/chains#5285


ligi (changes requested):
Provide the reason for the revert

PR #6042, opened 2024-10-22
Add Apechain network

joescherping:
This PR adds ApeChain mainnet (Chain ID: 33139)


joescherping:
closing because there is already a duplicate to add apechain mainnet

PR #6048, opened 2024-10-22
feat: add INCO testnet

0xshikhar:


0xshikhar:
Hey @ligi, the build is kept failing due to Cloudflare bot protection on explorer url ( explorer https://explorer.testnet.inco.org/ does not return 2XX but 403) - which is working when normally someone visits it.

ligi (approved):

PR #6058, opened 2024-10-23
feat: Add Inco Rivest Testnet

silasdavis:


Dubersnack (commented):
Overall, the data looks good, and the file is well-formatted. 

Here are a few minor suggestions:

* **`shortName`:**  Use "boba" as the `shortName`, as it seems to be the most common and recognized abbreviation.
* **`rpc`:**  It might be helpful to add a secondary RPC URL as a backup in case the first one experiences downtime.
* **`faucets`:** I noticed this is a mainnet chain.  It might be beneficial to remove the `faucets` field or explicitly state "N/A" to avoid confusion.

silasdavis:
@Dubersnack thank you for the swift review!

I have fixed the icon dimension typo from previous test failure.

> shortName: Use "boba" as the shortName, as it seems to be the most common and recognized abbreviation.

I'm sorry I don't recognise "boba", where are you getting this from? Open to suggestion on `shortName`, just wanted to capture that this is a testnet.

> faucets: I noticed this is a mainnet chain. It might be beneficial to remove the faucets field or explicitly state "N/A" to avoid confusion.

This is not a mainnet, it is a testnet, if that's what you mean, the faucet is necessary/functional right now.

> rpc: It might be helpful to add a secondary RPC URL as a backup in case the first one experiences downtime.

This is a centralised testnet the backups we have are behind a load-balancer, so for this testnet that's as good as it gets :)

Dubersnack (commented):
This commit adds support for the Rivest testnet by including its chain ID, RPC URLs, faucet, and explorer information. It also correctly sets the slip44 value to 1. The information appears accurate and complete.

ligi (approved):

PR #6062, opened 2024-10-24
Fixed content for Klyntar Mainnet+ aded new chainID for Klyntar Testnet

VladChernenko:
Hello, @ligi . Just make changes to URLs for RPC + add new network (testnet) for separation. Domains for explorers are valid but may respond with 500 code - but it still reachable. Once node will response for explorer queries - it will looks like in our [docs](https://docs.klyntar.org/build-core-and-join-network/explorers-and-how-to-use-them/your-own-custom-explorer)


VladChernenko:
Ok, once run - we'll repeat the build

PR #6066, opened 2024-10-24
Accumulate Kermit

firelizzard18:
Adds [Accumulate Kermit Testnet](https://kermit.explorer.accumulatenetwork.io/)


ligi (changes requested):
explorer does not seem to be EIP3091 compatible when it comes to addresses

firelizzard18:
@ligi I added a route for addresses: https://kermit.explorer.accumulatenetwork.io/address/0x77405B5aD47128c646D5D8fC5c2313c20F99B3Fc. It will still say "not found" if you put in an address that has never been used.

PR #6084, opened 2024-10-28
incubating swell mainnet 1923

cryptoquantx:
This is to reserve chainId 1923


tilemanrenovations:
> You successfully submitted a PR! Due to the amount of PRs coming in: we will only look at PRs that the CI is happy with. We can also not hold your hand getting the CI green - just look how others that where merged did it and RTFM. So as long as there is any CI check that reports an error - no human will look at this. You might be able to ask for some support after supporting the project - e.g. by sending funds to lists.eth. When you fixed things after a requested change - then you also need to (re-)request a review.

tilemanrenovations (commented):
TnMe4Rteam

ligi (changes requested):
add status incubating then and remove stuff that is not yet there like RPCs

ligi (approved):

PR #6099, opened 2024-10-30
Rebranding: eCredits to eSync Network

sieg-i:
- eCredits was rebranded to eSync Network. Previously it was only done for testnet, now also mainnet.

- added additional RPC endpoint with new esync.network domain

### DIFF
--- a/_data/chains/eip155-63000.json
+++ b/_data/chains/eip155-63000.json
@@ -23,7 +23,7 @@
     },
     {
       "name": "eSync Network Mainnet Beacon Explorer",
-      "url": "https://beacon-explorer.esync.network/",
+      "url": "https://beacon-explorer.esync.network",
       "icon": "esync",
       "standard": "EIP3091"
     }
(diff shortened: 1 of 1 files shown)


sieg-i:
- eCredits was rebranded to eSync Network. Previously it was only done for testnet, now also mainnet.
- added additional RPC endpoint with new esync.network domain

ligi (changes requested):
https://ecredits.com still exists

sieg-i:
> https://ecredits.com still exists

Not decided yet whether to redirect to https://esync.network or site will be rebuilt.

However block explorer is already set to redirect: https://explorer.ecredits.com to esync.network domain.

ligi (changes requested):
The beacon explorer is not EIP3091 compatible - don't even think it can be

sieg-i:
> The beacon explorer is not EIP3091 compatible - don't even think it can be

Should I remove the "standard" field from json then? Or drop the entire entry from explorers collection?

ligi (approved):

PR #6118, opened 2024-10-31
chore: add all lower nibiru envs

cgilbe27:


cgilbe27:
This MR is ready

ligi (changes requested):

cgilbe27:

Addressed

_data/chains/eip155-7210.json
```diff
@@ -8,10 +8,11 @@
...
     "decimals": 18
   },
+  "status": "deprecated",
   "infoURL": "https://nibiru.fi",
   "shortName": "nibiru-testnet-1",
-  "chainId": 7210,
-  "networkId": 7210,
+  "chainId": 6910,
+  "networkId": 6910,
   "icon": "nibiru",
```
Unique-Divine, line 16: Note that in other deprecated Chain ID files, it uses the deprecated "chainId" and "networkId", not the new values.

ligi (changes requested):
RPC chainId (7220) does not match chainId from json (6920)

cgilbe27:
> RPC chainId (7220) does not match chainId from json (6920)

This will eventually be our chain ID (6920)

PR #6120, opened 2024-10-31
Add Ink Sepolia

thINKoor:
Add Ink Sepolia


thINKoor:
I can fix validation? But would defer to you @ligi

ligi (approved):

PR #6155, opened 2024-11-06
Datacore Smart Chain

ankyjais:


ligi (changes requested):
quite confusing that name is still digital - can you elaborate?

ankyjais:
The name 'Digital Smart Chain' is causing some confusion. Could we update it to 'Datacore Smart Chain' on Chainlist to better reflect our brand?

ligi (approved):

PR #6160, opened 2024-11-06
Update eip155-205.json

allexandry:
change RPC to keleido


ligi (changes requested):
That looks sus to me - please elaborate and give proof of why this changes this way

PR #6163, opened 2024-11-07
Add my chain

ClearBlocker:


ligi (changes requested):
https://clearblocker.com does not look like a explorer to me

ClearBlocker:
There is an explorer you just need to be signed in to view it, I’ll setup a blockscout explorer though to satisfy the requirements here and then resubmit the request.

ligi (changes requested):
If you need to login it is clearly not EIP3091 compatible

ClearBlocker:
you no longer have to login to view the explorer it is live at https://explorer.clearblocker.com

PR #6172, opened 2024-11-08
Update bttc networks.

falcon202110:
Update BTTC Networks:

1. Add icon information for 199 network.
2. Add chain information for 1029 network.
3. Remove information for 1028 Network.


ligi (changes requested):

PR #6206, opened 2024-11-14
Create eip155-50311.json

Saksham-somnia:
Somnia DevNet - devnet.somnia.network


Saksham-somnia:
@ligi can you please check if everything here is good

PR #6165, opened 2024-11-07
add Betherance Chain ID

betheranceteam:

PR #6241, opened 2024-11-21
"add Betherance Chain ID"

betheranceteam:
Reverts ethereum-lists/chains#6165


EdiA1 (commented):
muy bueno

ligi (changes requested):

PR #6265, opened 2024-11-25
incubating UNITE

jtan-sl:
adding mainnet and testnet definitions


Goddessnunu22:
> adding mainnet and testnet definitions

ligi (approved):

PR #6282, opened 2024-11-27
feat: update corn maizenet and testnet

sajanrajdev:
Updated Corn Maizenet and Testnet's public RPC and official Block Explorer.


Attila180 (commented):
Hi, I learned

ligi (approved):

PR #6320, opened 2024-12-03
Changed mainnet name to Lux Mainnet

zoosos:


ligi (changes requested):
No change -> no merge

PR #6336, opened 2024-12-04
Updated Kaia Testnet rpc endpoint

praveen-kaia:
Updated Testnet RPC url as per below Kaia docs
https://docs.kaia.io/references/public-en/#testnet-kairos-public-json-rpc-endpoints


praveen-kaia:
Thanks @ligi . Appreciate your quick response.

PR #6339, opened 2024-12-04
Update eip155-7082400.json

gmesika-coti:
added more COTI validators


gmesika-coti:
will be fixed in a seperate PR

PR #6349, opened 2024-12-05
Add AppChain 

emrme:


emrme:
@ligi

emrme:
@ligi updated the icon size, the `validate_icons` check was failing

ligi (approved):

PR #6353, opened 2024-12-05
Revert "Update eip155-50104.json"

Eikyarkway:
Reverts ethereum-lists/chains#6348


ligi (changes requested):
why the revert?

ligi (changes requested):
you cannot request a review without answering my question

PR #6374, opened 2024-12-08
Corrected the explorer url 

Tixandre:
We change the explorer url to be the mainnet and testnet on same explorer


ligi (changes requested):
the /NTT seems to make it incompatible with EIP3091

PR #6406, opened 2024-12-13
change chain name from jzero to jfinpos

dome:
change chain name from jzero to jfinpos


ligi (changes requested):
https://jfinpos.com is not reachable

dome:
> https://jfinpos.com is not reachable

fixed

ligi (approved):

PR #6489, opened 2024-12-21
Update Hydra Testnet faucet and RPC URLs because of DNS issues

R-Santev:
The PR includes an update of the Testnet faucet and rpc URLs that has to be made because of DNS problems


Sajjadshegeft:
This project is very good and important

ligi (approved):

PR #6518, opened 2024-12-30
Add Nexera Testnet + Mainnet

galimba:
Adding Nexera testnet (chain id `72080`) + mainnet (chain id `7208`)


galimba:
@ligi  :pray:  now ready to review, I think

ligi (approved):

PR #6569, opened 2025-01-09
Add Ultra EVM Testnet

Duncan-Ultra:
Add Ultra EVM Testnet with ChainId 18881


Duncan-Ultra:
It fails because of the block explorer address.
That address is not live yet since our team will be deploying it in the next few weeks, so it's just a placeholder.

I will be back when the explorer are live.

Duncan-Ultra:
@ligi we have mapped a DNS to the URL, could you run the check again?

PR #6596, opened 2025-01-15
update Soonchain Devnet Sepolia info

AphobiaCat:
update Soonchain Devnet Sepolia explorer info


_data/chains/eip155-20221001.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "SoonChain Sepolia Devnet",
+  "name": "SoonChain Devnet Sepolia",
   "chain": "SoonChain Devnet Sepolia",
   "rpc": ["https://sepolia.rpc.soonchain.ai"],
-  "faucets": ["https://console.optimism.io/faucet"],
+  "faucets": ["https://docs.optimism.io/builders/tools/build/faucets"],
```
ligi, line 5: elaborate this change

ligi (changes requested):

_data/chains/eip155-20221001.json
AphobiaCat: The original link for this faucet was only developer friendly, but in fact it was more difficult for users to pick up, so we switched to the list of faucets included in the op Stack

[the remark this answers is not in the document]
AphobiaCat: okay, Then there is no need for additional links

PR #6615, opened 2025-01-17
Add Uxlink One  Mainnet  chain id

MarkStephon:


_data/chains/eip155-718.json
```diff
@@ -0,0 +1,25 @@
...
+  "features": [{ "name": "EIP155" }, { "name": "EIP1559" }],
+  "infoURL": "https://www.uxlinkone.io",
+  "shortName": "uxlink1",
+  "chainId": 718,
+  "networkId": 718,
+  "icon": "uxlinkone",
+  "explorers": [
+    {
+      "name": "UXLINK ONE Mainnet Explorer",
+      "url": "https://www.uxlinkone.com",
```
ligi, line 20: this does not look like an explorer - especially not a EIP3091 compatible one - at least I cannot verify like that

ligi (changes requested):

PR #6649, opened 2025-01-24
update settlus info

devbv:
As we announced the launch of the Settlus Sepolia Testnet, we’ve updated information about Settlus.


ligi (approved):

Rias0232022:
Ok

PR #6661, opened 2025-01-26
add merlin erigon testnet

lyh169:


lyh169:
> https://testnet-erigon-scan.merlinchain.io
> 
> does not seem to be an explorer
it will be better after a few days

PR #6667, opened 2025-01-27
updated Haust Network Testnet

ogunleye0720:
Removed "status" field from Haust Network Testnet


ogunleye0720:
Yeah @ligi was only wondering why it wasn't showing up. Thanks a lot 🙂

PR #6670, opened 2025-01-27
Create eip155-2712666325747000.json

jonathan366:
Add Riftwalkers Chain


ligi (changes requested):
the Icon riftwalkers does not exist - was used in eip155-2712666325747000.json

Rias0232022:
Ok

PR #6709, opened 2025-02-01
chore: deprecate nibiru 7210 chain env

cgilbe27:
Deprecates 7210 and adds 6910


k-yang:
Chain ID 6910 is live and running for our testnet

cgilbe27:
@ligi Following up on this

ligi (changes requested):

k-yang:
Hi @ligi , we marked it as deprecated but then the build CI failed because it was expecting the old chainId (7210). What's the process for changing the chainId of a test network?

k-yang:
Hi @ligi, please advise

ligi (changes requested):
https://github.com/ethereum-lists/chains/pull/6709#issuecomment-2629035636

PR #6718, opened 2025-02-03
Updated GPT mainnet chainID

arpitkarnatak:


ligi (changes requested):

arpitkarnatak:
@ligi Thank you for highlighting this. We've moved our chain to a new ChainID (111222333) and it's running on the same URLs as our old ChainID (1511670449). However the CI is failing for the deprecated chain ID (1511670449) since the ChainID doesn't match the chainID from the RPC (we don't have an RPC with the old chain ID). Please let me know what needs to be done for this situation, thank you and have a good day

ligi:
the solution is quite obvious - but I cannot give free support - send funds to lists.eth and I let you know

PR #6726, opened 2025-02-03
Added Rebus L2 configuration and updated Rebus Classic

nicolam1:
Added configuration for Rebus Layer 2 and updated existing Rebus to Classic


nicolam1:
@ligi is it possible to merge the PR so we can launch our layer 2 🙏

ligi (approved):

PR #6783, opened 2025-02-11
Update to GH Actions V4

trianglesphere:
Github is currently doing a brownout period for cache V2 which is causing PR checks to fail.
https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down


ligi (approved):
thanks!

PR #6802, opened 2025-02-14
Create eip155-2022091 Alterscope Testnet

zvedran:
Alterscope Testnet eip155-2022091


zvedran:
Is there an issue with gradle?

PR #6838, opened 2025-02-19
update dis to ethf

FranckStone:
update dis to ethf


ligi (changes requested):
Clarify if this is a rename of a chain or if there is a new genesis

FranckStone:
this is a rename

ligi (approved):

PR #6847, opened 2025-02-20
[New Chain] Recall testnet and mainnet

oed:


ligi (changes requested):
image

oed:
@ligi yes the mainnet isn't live yet. I can remove the link and add it later if that is better?

ligi (changes requested):
yes - remove for now then - I cannot merge like this as I also cannot verify it is EIP3091 compliant.

_data/chains/eip155-2481632.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "Recall Testnet",
+  "chain": "Recall Testnet",
+  "rpc": [" https://evm.v013.node-0.testnet.recall.network"],
```
ligi, line 4:
```suggestion
  "rpc": ["https://evm.v013.node-0.testnet.recall.network"],
```

ligi (changes requested):

_data/chains/eip155-2481632.json
oed: 🤦

PR #6851, opened 2025-02-20
fix: 🐛 Deprecated Artio

codingwithmanny:


codingwithmanny:
Hi @ligi could you help advise how to handle this?

The only solution I found was the following, but requires some gradle configurations:
```json
allprojects {
    tasks.withType(JavaCompile) {
        options.deprecation = true
    }
}
```
Reference:
https://code2care.org/gradle/fix-deprecated-gradle-features-were-used-in-the-build-incompatible/

ligi (changes requested):
RPC chainId (1) does not match chainId from json (80085)

-> remove the RPCs

codingwithmanny:
> RPC chainId (1) does not match chainId from json (80085)
> 
> -> remove the RPCs

Much appreciated @ligi

ligi (approved):

PR #6862, opened 2025-02-23
edeXa mainnet update

aseempm:


ligi (changes requested):
why do you remove the features?

aseempm:
To make as similar as our Testnet Configuration in ChainList.

Testnet: 1995

ligi:
but don't your chain has these features?

aseempm:
Yes. We do have. !

ligi (approved):

PR #6871, opened 2025-02-24
Update BXN chain

BlackFortGroup:


BlackFortGroup:
Hello @ligi, our updated chains remain same symbol name, that's the reason of failing build. Could you assist us?

ligi (changes requested):
short name must be unique - but found `BXN` more than once

ligi (approved):

PR #6885, opened 2025-02-25
Update Electroneum (ETN) mainnet & testnet rpc

andrepatta:


andrepatta:
@ligi done. I've also updated the testnet block explorer url.

Thanks

PR #6913, opened 2025-02-27
feat: add xchain testnet and mainnet

dileepfrog:


_data/chains/eip155-64002.json
Morhef, line 1: _data/chains/eip155-94524.json

Morhef (commented):
_data/chains/eip155-94524.json

ligi (changes requested):

dileepfrog:
Added deprecation notice to old xchain listings - this unrelated project appears defunct as none of the provided URLs work. Added `deprecated` string where applicable to avoid naming collision. Please advise what else we can do to get this merged, thank you!

dileepfrog:
The build error is for the old chain that we flagged deprecated:

`Exception in thread "main" java.net.UnknownHostException: xchainscan.com: No address associated with hostname`

ligi (changes requested):
remove these URLS then

dileepfrog:
> remove these URLS then

Removed

ligi (approved):

PR #6917, opened 2025-02-28
Added Primea Chain ID 1698369

primeanetwork:
Added Primea Chain Testnet 1698369
changed Primea Chain name to Primea Chain Mainnet


DUMIE5050 (commented):
Review

PR #6934, opened 2025-03-03
Add Juneo mainnet chain information and logo

falcoda:
New files added : 

- Juneo mainnet chain information
- Juneo logo

### DIFF
--- a/_data/chains/eip155-45003.json
+++ b/_data/chains/eip155-45003.json
@@ -1,0 +1,25 @@
+{
+  "name": "Juneo Mainnet",
+  "chain": "JUNE-Chain",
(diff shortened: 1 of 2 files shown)


ligi (changes requested):
explorer does not look EIP3091 compatible

falcoda:
> explorer does not look EIP3091 compatible

Hello, I only found that we have blocks instead of block in this url ( `<BLOCK_EXPLORER_URL>/block/<BLOCK_HASH_OR_HEIGHT>` ). And also I'm not sure if have an ability to query block with it's height.

I will update the json file without EIP3091 and update it later when we will be fully compatible

ligi (approved):

PR #6967, opened 2025-03-07
add new Plume mainnet

eyqs:
Users on the existing Plume private mainnet with chain ID 98865 will be migrated over to the new Plume private mainnet over the next several weeks, before we mark the existing Plume private mainnet as deprecated.


_data/chains/eip155-98865.json
Morhef, line 1: 😁

_data/chains/eip155-98866.json
Morhef, line 1: 😍

ligi (approved):

PR #6980, opened 2025-03-10
inertia scan

arpitjais1:


ligi (changes requested):
do not check in package-lock.json

arpitjais1:
Kindly please merged the inertia scan

ligi (changes requested):
no change -> no merge

PR #6983, opened 2025-03-10
Add Xphere Mainnet

Kim-DaeHan:
Request PR to add Xphere mainnet.


ligi (changes requested):

Kim-DaeHan:

I checked your request and deleted Explorer

PR #6989, opened 2025-03-11
Updated to Nexurachain

dwinea:


ligi (changes requested):
You need to clarify: is this a chain rename?

dwinea:
Yes this is chain rename. From Magichain to Nexurachain 
@ligi

ligi (approved):

PR #7012, opened 2025-03-14
Ultima and Smart blockchain data added

smartdev2023:


ligi (changes requested):
do not check in this file:
image

PR #7038, opened 2025-03-17
Feat: add hoodi

mcdee:


ligi (changes requested):
hoodi is already in the list - and I do not think the changes are really good

mcdee:
Sorry it was missing when I started the changes.  Will cose.

PR #7045, opened 2025-03-19
add Load Network chain and icon

xylophonez:
WeaveVM rebranded to Load Network (https://blog.load.network/weavevm-is-now-load-network/). 

This PR updates the 9496 metadata and adds a new loadnetwork icon.


ligi (approved):
It seems to just be a rename

PR #7066, opened 2025-03-23
Add Fuga Network

fuga-one:
Add Fuga Network

### DIFF
--- a/_data/chains/eip155-1732.json
+++ b/_data/chains/eip155-1732.json
@@ -1,0 +1,23 @@
+{
+  "name": "Fuga Mainnet",
+  "chain": "Fuga",
+  "rpc": ["https://rpc.fuga.blue"],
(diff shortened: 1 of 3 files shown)


ligi (changes requested):
The explorer https://explorer.fuga.blue does not show anything

fuga-one:
Hi, ligi.

The block explorer is currently under construction and is not yet available for viewing. 
This PR is submitted solely to reserve the ChainID. 
Would it be possible to register only the chainId before the block explorer is built?

ligi:
yea - remove the explorers and add status incubating

fuga-one:
Thank you for your guidance. 
I have now removed the explorers value and added the status field.
Could you please review it again ?

fuga-one:
I updated the source code again, so please check it.

PR #7088, opened 2025-03-26
add Xorai Zathura Testnet Chainid 7973

xorainet:
add Xorai Zathura Testnet Chainid 7973


xorainet:
please update chain iD

PR #7115, opened 2025-03-30
Create prittier

illuminated-shade:
[Run prettier check](https://github.com/ethereum-lists/chains/actions/workflows/prettier_check.yml)


ligi (changes requested):
can you elaborate what this does?

PR #7178, opened 2025-04-12
Add deinfra devnet

cleverfox:
add deinfra devnet


cleverfox:
@ligi @FrederikBolding Could you check this PR please?

PR #7192, opened 2025-04-15
Create eip-155-7131.json

khuramShabbir:
add new chain VRCN Chain-7131


khuramShabbir:
@ligi 
Waiting for your approvel

PR #7207, opened 2025-04-21
Remove alaya mainnet and alayadev network

benbaley:
[Announcement](https://www.alaya.network/)


benbaley:
Hi @ligi please take a look at this, I just remove 3 files about Alaya network, why it's built failed? How can i fix this please

PR #7265, opened 2025-05-06
 Remove Neonscan as default explorer from Neon EVM Devnet and Mainnet chains

sukanyaparashar:
This PR is from Neon EVM's team to remove the native Neonscan explorer from the Neon EVM Devnet and Mainnet chain configurations.


ligi (changes requested):
why the removal?
can you prove you are part of the team?

sukanyaparashar:
> why the removal? can you prove you are part of the team?

Hi, Thanks for looking into this. You can see from my profile https://github.com/sukanyaparashar that I am part of [neonlabsorgs](https://github.com/neonlabsorg).

We are migrating to Blockscout as our default block explorer and our team has decided to remove Neonscan completely and keep only Blockscout as our explorer.

PR #7269, opened 2025-05-07
Mezo mainnet config

pdyraga:
Mezo mainnet chain configuration with an icon. 

I have also refreshed the Mezo testnet config to remove the no longer available faucet and to reflect the name change. From now on, it is just "testnet".


pdyraga:
Hey @FrederikBolding @ligi, pardon the direct ping, but we are hoping to launch Mezo mainnet in May, and this PR [is a blocker](https://github.com/safe-global/safe-singleton-factory?tab=readme-ov-file#how-to-get-the-singleton-deployed-to-your-network) for deploying the Safe singleton factory. Any chance we could safe-track it, please?

_data/chains/eip155-31612.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "Mezo",
+  "chain": "Mezo",
```
FrederikBolding, line 3: Duplicated chain name between mainnet and testnet.

Also seemingly missing `icon`?

pdyraga: Fixing now.

_data/chains/eip155-31612.json
```diff
@@ -0,0 +1,30 @@
...
+  ],
+  "faucets": [],
+  "nativeCurrency": {
+    "name": "Bitcoin",
+    "symbol": "BTC",
+    "decimals": 18
+  },
+  "features": [{ "name": "EIP155" }, { "name": "EIP1559" }],
+  "infoURL": "https://mezo.org/",
+  "shortName": "mezo",
```
FrederikBolding, line 18: Shortname also duplicated

pdyraga: Fixed in 7d976e822f56be9d341899cda7873e6ddc0367ce. I realized your comment was about it at the same moment I pushed the commit. 🙂 It feels like the `shortName` has to be unique, but `chain` doesn't have to, right? At least this is what I get looking at other examples, like 80094 and 80069.

Anyway, probably better to keep them all different to avoid any confusion.

FrederikBolding (approved):

PR #7339, opened 2025-05-23
feat: add Reddio Mainnet 

haisenCx:
This PR configures Reddio's mainnet.


haisenCx:
> _ I understand nothing_

sry ser, this is for reddio mainnet config

haisenCx:
@LordGoda  Sorry to direct ping, could you please take another look? We have some time-sensitive partnerships requiring this.

PR #7344, opened 2025-05-26
add note-main chain to chainlist

note-design:


note-design:
We hope to release the note-main chain tomorrow,thanks

PR #7430, opened 2025-06-17
Update Plume's name

silvestreh:
This PR updates Plume's mainnet name so that it matches that of [Viem](https://github.com/wevm/viem/blob/main/src/chains/definitions/plumeMainnet.ts).


silvestreh:
Can anyone take a look at this?

PR #7439, opened 2025-06-19
feat:add qubetics chain list

NodeNavigator:
add qubetics chain list


NodeNavigator:
Hi Sir, when will this branch be merged? We have a mainnet launch on 1st July.

PR #7456, opened 2025-06-23
fix: resolve deployment error with loading components and Suspense

Dhy2928:
Add required loading components and Suspense boundaries for better UX.


.ci/kontinuum.json
Dhy2928, line 1: Approve

PR #7472, opened 2025-06-27
Add chainova-CNV

Daimon-Security:
https://chain-nova.tech


Daimon-Security:
@artob @farcaller  @zapnap  pls check

PR #7501, opened 2025-07-08
Pr/9030 mainnetqubetics

NodeNavigator:
add qubetics chain


NodeNavigator:
Hi, please merge this request. Our chain is already live, and we need to register it on the Ethereum Chainlist to ensure our chain ID is not used by others. This will also allow our token price to appear in the MetaMask wallet and let users see that our chain is officially listed on the ETH Chainlist.

PR #7510, opened 2025-07-10
aipaw icon

billpi1:
Adding Aipaw chain 90025 with 1024x1024 icon to Chainlist.org.


billpi1:
@FrederikBolding please add it

PR #7554, opened 2025-07-29
Shinerium eip155-0505 chain with icon

shin-usagi:


shin-usagi:
I submitted this PR and ran local validation via `cradle`, which passed.

PR #7555, opened 2025-07-29
Add shinerium chain  icon

shin-usagi:
shinerium icon


shin-usagi:
Closed as automatic checks seems not to work with this

PR #7606, opened 2025-08-12
feat: update mocachain devnet url

athaswang1991:
We update a new url for mocachain devnet. Pls appr!


athaswang1991:
Pls process it asap.

PR #7684, opened 2025-09-11
Create eip155-9601.json

dguang:
include RPC, explorer, and that EIP-155/EIP-1559 are supported.


dguang:
could a maintainer please approve running the workflows? Thanks!

ligi (changes requested):
do not see transactions/blocks in the explorer and hence cannot verify EIP3091 compatibility

PR #7697, opened 2025-09-15
Add Lightning Vault chain (chainId 296070)

gmcradz:


gmcradz:
Name: Lightning Vault
RPC: https://lightning-mainnet.site
Explorer in the works
TICKER: LVT

PR #7725, opened 2025-09-21
Update Jovay mainnet network config

zoowii:
Update Jovay network default public rpc endpoint

Hello Team!

I've updated the data for Jovay Mainnet.

Changes:

Updated the default rpc endpoint for the Jovay Mainnet network.

### DIFF
--- a/_data/chains/eip155-5734951.json
+++ b/_data/chains/eip155-5734951.json
@@ -3,7 +3,7 @@
   "chain": "ETH",
   "status": "incubating",
   "rpc": [
-    "https://api.zan.top/public/jovay-mainnet",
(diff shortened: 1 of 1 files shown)


zoowii:
@FrederikBolding Jovay's mainnet has launched, so we need to modify the configuration. Please review it when you have time 🙏

FrederikBolding:
@zoowii looks good, but seems to fail CI?

zoowii:
It's strange, the CI failure logs don't show which specific URL in the json failed. I checked, and the curl request to the new RPC endpoint is the same as before. I can also curl the block explorer and the official website. I rebased (but I don't think that caused the issue). Should we try re-running the CI?

PR #7787, opened 2025-10-15
add cycle mainnet frigate

max2024666:
add cycle mainnet frigate with chain id 88688


max2024666:
@FrederikBolding would you please check this PR ? thanks

PR #7795, opened 2025-10-20
Add Tensora L2 (Chain ID: 44444444)

TensoraL2:
This PR adds Tensora L2 to the official chain list.
  
  - Chain ID: 44444444
  - Parent: BSC (eip155-56)
  - RPC: https://rpc.tensora.sh
  - Explorer: https://explorer.tensora.sh
  - Status: active


jayjose1:
can we run validator now

PR #6840, opened 2025-02-19
updated Hiperliquid Chain

alfredolopez80:
Update Data, logo, chain info and RPC of Chain Id: 999, that corresponding to Hyperliquid EVM L1: more details: https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/evm

PR #7810, opened 2025-10-27
Add HyperEVM

hl-core-dev:


hl-core-dev:
WanChain testnet appears to be effectively deprecated. For example, the explorer is not functional. 

But open to suggestions on alternatives for adding HyperEVM to the list.

FrederikBolding:
The stance on HyperEVM is the same stance as any other chain that tries to replace a non-deprecated chain.

https://github.com/ethereum-lists/chains/pull/6840#pullrequestreview-2678248622

https://github.com/ethereum-lists/chains?tab=readme-ov-file#collision-management

jxom:
> WanChain testnet appears to be effectively deprecated.

Even though the chain is deprecated, users are still prone to replay attacks.

It's only a matter of time before a previous user of wanchain will get replay attacked on HyperEVM. :)

PR #7813, opened 2025-10-27
Update rollux testnet 20250703

fernando-syslabs:


_data/chains/eip155-57000.json
```diff
@@ -1,29 +0,0 @@
-{
-  "name": "Rollux Testnet",
```
FrederikBolding: Please mark as deprecated instead of deleting

PR #7814, opened 2025-10-27
Remove Syscoin Rollux Testnet

fernando-syslabs:
Rollux Testnet has been officially discontinued: https://syscoin.org/news/ecosystem-update-251016


ligi (changes requested):

PR #7836, opened 2025-11-03
Add fraxtal hoodi testnet - ready to merge

0xalex88:
This commit adds the new Fraxtal testnet based on hoodi and deprecates the old testnet (which is already down).


0xalex88:
@ligi I've removed our old explorer as it has been decommissioned

0xalex88:
Thank you @ligi !

PR #7843, opened 2025-11-05
Create GBTNetwork 09636.json

alexwambu:
Add GBTNetwork chain id  9636


alexwambu:
Add GBTNetwork chain id 9636

ligi (changes requested):
file in wrong place

PR #7917, opened 2025-11-30
Add Kalichain Subnet (chainId 6533) and deprecate Kalichain (chainId 654)

KalissaIO:
Closing this PR in favor of #8022 which removes all Kalichain data instead of deprecating and adding new chains.


ligi (changes requested):
please do no touch package-lock.json

PR #7989, opened 2026-01-14
Adds Polkadot EVM details

albertov19:
Adds Polkadot EVM chain details


albertov19:

Apologies, makes sense. fixed. (I did not mark Westend as deprecated, as it is not, but just "unstable" for developers)

albertov19:
Hey @ligi let me know if anything else 🙏  - Polkadot EVM is going live tomorrow. Thanks in advance

PR #7990, opened 2026-01-15
Fix HashKey Chain Testnet infos

fortyfourZ:


ligi (changes requested):
Image

also I need some kind of proof this change is wanted by the chain owners

fortyfourZ:
> Image
> also I need some kind of proof this change is wanted by the chain owners

in official chain account [https://x.com/HSKChain], there are links for HashKey Chain. And the Official Website is https://www.hashkeychain.net.

PR #8000, opened 2026-01-20
Add KiiChain mainnet data

GiovaniGuizzo:
* Adds the chain data for KiiChain mainnet
* Currently private mainnet, but this will make it possible for us to go public


GiovaniGuizzo:
@ligi I just updated the branch to be in sync with your "master". Could you approve again please? :)

PR #7361, opened 2025-05-29
Add bokuto (katana testnet) (737373)

ToniRamirezM:
This PR adds bokuto (katana testnet).

PR #8006, opened 2026-01-22
add Bokuto (current Katana Testnet)

omnifient:


omnifient:
btw, I see that another chain has an opened PR for 737373, but I want to note that we tried to add this 2x before

- May 29th https://github.com/ethereum-lists/chains/pull/7361 
- October 17th https://github.com/ethereum-lists/chains/pull/7794

the other chain's PR was opened in June, so after we had our 1st PR pending

PR #8007, opened 2026-01-23
Modify X Layer Tetstnet

giskook:
# Chain Name

X Layer Testnet
# Description 

We have officially sunset the X Layer Testnet and it is no longer operational. We would like to request an update to the network listing to reflect its decommissioned status, or remove it from the available testnet networks if appropriate.

Thank you for your assistance.


ligi (changes requested):

giskook:
Hello @ligi, Thanks for your feedback, I've modified it, pls take a look :)

giskook:
Hi @ligi , We reused the 195's RPC URL, do we need to use an URL alias？ eg https://testrpc.xlayer.tech/xlayer etc? Many thanks.

ligi:
you can remove all URLs - just not delete the chain

giskook:
@ligi I removed all URLs, pls check it, Many thanks

PR #8018, opened 2026-01-26
deprecate tatara (old katana testnet)

omnifient:


omnifient:
hi team (@ligi), the build seems to be failing on this check:

> Exception in thread "main" javax.net.ssl.SSLPeerUnverifiedException: Hostname explorer.tatara.katana.network not verified:
>     certificate: sha256/J34h4O1fUHdJRCPTjO0YrTIsbS4yjIEHonjBQVhNIKs=
>     DN: CN=*.t.conduit.xyz
>     subjectAltNames: [*.t.conduit.xyz]

which is caused by the fact that the network has been shutdown - this PR is to make it deprecated, so can we manually bypass this? 

tyvm

omnifient:
bumping for visibility

PR #8032, opened 2026-01-30
Update CratD2C testnet info

joxerx:
- update testnet rpc list
- update testnet chainId


ligi (changes requested):

joxerx:

thanks for your time. marked previous file as deprecated

PR #8134, opened 2026-03-13
Remove the unused PlatON testnet2

benbaley:
## Summary
Remove the unused **PlatON Devnet2** (Chain ID: 2206132)

## Files removed
* `_data/chains/eip155-2206132.json`

## Validation
* ✅ Schema check passed
* ✅ Formatted with Prettier
* ✅ Icon accessible via IPFS
* ✅ No conflicts with existing chains


ligi (changes requested):

PR #8173, opened 2026-04-01
Remove Cyber Testnet (chain ID 111557560)

PengDeng-Cyber:
## Summary
- Remove the Cyber Testnet chain (chain ID 111557560) as the network has been shut down and is no longer operational.

## Details
- File removed: `_data/chains/eip155-111557560.json`
- The Cyber Mainnet (chain ID 7560) remains unaffected.


ligi (changes requested):

PR #8192, opened 2026-04-07
chore: update the RPC endpoints, explorer URL, official website URL for the CarrChain Network

artiffine-vojtech:
Update info for mainnet chain ID 7667

Update info for testnet chain ID 76672

Mark previous testnet chain 77652 as deprecated.


artiffine-vojtech:
Hi @ligi, quick question. Since the old and new testnets share the same chain short name, would you suggest renaming the old chain or assigning a new name to the new testnet?

The error from CI pipeline:
`Exception in thread "main" org.ethereum.lists.chains.model.ShortNameMustBeUnique: short name must be unique - but found `CARRCHAIN-TESTNET` more than once`

artiffine-vojtech:
Hi @ligi I renamed the short name of the old deprecated chain. Let me know if this is the right approach. Thanks!

artiffine-vojtech:
Hi, how we can proceed with this PR to have it merged? 
cc @ligi 

Thanks

ligi (approved):

PR #8302, opened 2026-05-08
Feature/add chain 787111

BigtoMantraDev:


BigtoMantraDev:
Build error is expected, because our native symbol is `MANTRAUSD`

```
Exception in thread "main" org.ethereum.lists.chains.model.NativeCurrencySymbolMustHaveLessThan7Chars: Native currency symbol must have less than 7 chars
	at org.ethereum.lists.chains.MainKt.checkChain(Main.kt:302)
	at org.ethereum.lists.chains.MainKt.main(Main.kt:44)
```

PR #8315, opened 2026-05-11
Migrate TreasureNet Aurora from chainId 5002 to 5570

chriszhao1988:
## Summary
- deprecate the legacy TreasureNet Aurora chain entry on chainId 5002
- add the canonical TreasureNet Aurora chain entry on chainId 5570
- update the RPC to https://rpc.treasurenet.io and explorer to https://scan.treasurenet.io

## Context
Chainlist and other downstream consumers ingest data from this registry. TreasureNet Aurora should use chainId 5570 instead of the old 5002 value, while the old entry remains preserved as deprecated metadata.


ligi (commented):
does 5002 have the same genesis

PR #8332, opened 2026-05-13
890 Discontinued

capitalexchange:


ligi (changes requested):
set the status to deprecated

capitalexchange:
> set the status to deprecated

In file the status is set to: status": "deprecated"

ligi (approved):

PR #8333, opened 2026-05-13
Add EIP155-343 chain data for Capital Exchange SE

capitalexchange:


ligi (changes requested):

capitalexchange:

Correct. Should've been "none".

ligi (approved):

PR #8354, opened 2026-05-19
Add DimeX Chain configuration (Chain id 1601)

glomartcom-netizen:


ligi (changes requested):
explorer is not EIP3091 compatible

PR #8357, opened 2026-05-19
Add ChainID 540 for private enterprise network

tiberi00:
Reserving ChainID 540 for a private Hyperledger Besu deployment to prevent future collisions. No public RPC endpoints are exposed for security reasons.


tiberi00:
updated json file

tiberi00:
inserted a localhost rpc

ligi (changes requested):
remove the localhost RPC - this is BS

ligi (approved):

PR #8392, opened 2026-05-30
Update Syscoin Mainnet icon and RPCs

sidhujag:
## Summary
- Add the Syscoin icon metadata and downloaded SVG asset for Syscoin Mainnet.
- Reference the icon from `eip155-57.json`.
- Remove `https://syscoin.public-rpc.com`, which currently returns an unauthenticated Ankr error instead of a chain ID.

## Validation
- `node schemaCheck.js`
- `npx prettier --check '_data/chains/eip155-57.json' '_data/icons/syscoin.json'`
- Verified the icon payload hashes to `bafybeidr6pfhdbtoafs5otqop6d2hau5zep4xvkpu2bpzjfgvbf6lzvnoa`.
- Checked remaining Syscoin RPCs return chain ID `0x39` where public access is expected.

Related context: #8391 adds Tanenbaum entries and uses the same Syscoin icon asset for Syscoin Tanenbaum Testnet.


sidhujag:
Closing this for now so it does not duplicate the shared Syscoin icon files already introduced in #8391. Once #8391 lands, this can be reopened/rebased as a small follow-up that only updates Syscoin Mainnet metadata.

PR #8429, opened 2026-06-13
Fix MFENX RPC endpoint for chain 177155

JROChub:
The existing MFENX RPC URL does not resolve. This replaces it with the live canonical endpoint behind the health-checked multi-region edge.

Validation:
- ChainList single-chain Gradle check passed
- Endpoint returned `power-house/0.3.2/finalized-native-rpc`
- Chain ID `177155` and block number were validated by the ChainList processor
- Prettier and JSON syntax checks passed

### DIFF
--- a/_data/chains/eip155-177155.json
+++ b/_data/chains/eip155-177155.json
@@ -1,7 +1,7 @@
 {
   "name": "mfenx",
   "chain": "MFENX",
-  "rpc": ["https://indexer.mfenx.com/rpc"],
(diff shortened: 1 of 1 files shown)


JROChub:
Ready for maintainer merge queue. All upstream checks pass, the ChainList single-chain validator connects successfully, and https://rpc.mfenx.com is live on chain ID 177155 with client power-house/0.3.2/finalized-native-rpc.

ligi (approved):

PR #8490, opened 2026-07-08
Update details for EDU Chain mainnet and testnet

m4ttw4n:
Hi team, updating a few of the URLs for both EDU Chain mainnet and testnet.


ligi (changes requested):
Please add proof for these changes - e.g. mentioning the new URLs behind the old ones

PR #8500, opened 2026-07-10
Add NodeFlare public RPC endpoints for 15 chains

nodeflareapp:
Adds NodeFlare's free public RPC endpoints (`https://rpc.nodeflare.app/{chain}/public`) to 15 chains: Unichain (130), Sonic (146), Polygon (137), Linea (59144), Mantle (5000), Zircuit (48900), XLayer (196), Soneium (1868), Arbitrum Nova (42170), BOB (60808), Ink (57073), Cronos (25), Mode (34443), Sei (1329), Plasma (9745).

All endpoints are live and verified — correct `eth_chainId`, synced to chain tip, no API key required, served from multi-region bare-metal nodes with automatic failover. Same provider as previously merged #8359.

PR #8503, opened 2026-07-11
Add NodeFlare public RPC to 4 under-served chains (Plasma, Soneium, Zircuit, Sei)

nodeflareapp:
Adds NodeFlare's public RPC endpoint to four chains that currently list only a single HTTPS RPC, giving builders a fallback if the primary endpoint is down.

| Chain | File | Existing HTTPS RPC | Added |
|---|---|---|---|
| Plasma Mainnet | eip155-9745 | 1 | `https://rpc.nodeflare.app/plasma/public` |
| Soneium | eip155-1868 | 1 | `https://rpc.nodeflare.app/soneium/public` |
| Zircuit | eip155-48900 | 1 | `https://rpc.nodeflare.app/zircuit/public` |
| Sei Network | eip155-1329 | 1 | `https://rpc.nodeflare.app/sei/public` |

Each endpoint returns the correct `eth_chainId` and serves standard JSON-RPC with no key required. NodeFlare runs its own dedicated nodes for these chains across multiple regions with automatic failover.

Kept the diff to a single added URL per file; no other fields touched.

### DIFF
--- a/_data/chains/eip155-1329.json
+++ b/_data/chains/eip155-1329.json
@@ -1,7 +1,7 @@
 {
   "name": "Sei Network",
   "chain": "Sei",
-  "rpc": ["https://evm-rpc.sei-apis.com", "wss://evm-ws.sei-apis.com"],
+  "rpc": ["https://evm-rpc.sei-apis.com", "wss://evm-ws.sei-apis.com", "https://rpc.nodeflare.app/sei/public"],
   "faucets": [],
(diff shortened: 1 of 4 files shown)


nodeflareapp:
Closing: the 4 chains here are included in #8500 (just refreshed onto current master), so keeping one PR avoids the two conflicting on the same files. All endpoints remain live and verified.

PR #8564, opened 2026-08-03
Add Robinscan explorer for Robinhood Chain (eip155-4663)

robinscan:
Adds [Robinscan](https://robinscan.io) as a block explorer for Robinhood Chain (chain ID 4663), alongside the existing Blockscout explorer.


robinscan:
@ligi thanks for updating the branch — quick heads-up: the `build` failure that came with it is unrelated to this PR's changes.

`_data/chains/eip155-4663.json` also lists `https://hoodscan.ai` (added in #8512), which has been returning 402 with `x-vercel-error: DEPLOYMENT_DISABLED` for more than a week. That's not a transient error — the deployment was switched off, so the explorer check will keep failing for any PR touching this chain file, not only this one.

Since we're adding an explorer for the same chain, we'd rather not remove another project's entry unprompted — so we've opened #8601 with the removal, for you to merge or close as you prefer. It can always be re-added if the site comes back.

ligi (approved):

PR #8578, opened 2026-08-06
Reassign deprecated Chain ID 890 to MarsChain Mainnet

zqdd-iillo:
## Reassign deprecated Chain ID 890 to MarsChain Mainnet

The previous network registered under Chain ID 890, Capital Exchange, was officially marked as discontinued and deprecated through PR #8332:

https://github.com/ethereum-lists/chains/pull/8332

That PR was submitted by the original Capital Exchange project account and merged into the `master` branch in commit `1916a4e87e19a5c17c285fb0913fe1ae914a650c`.

MarsChain is an independent, active EVM-compatible Layer 1 mainnet that currently operates using Chain ID 890.

### MarsChain information

- Network name: MarsChain Mainnet
- Chain ID: 890
- Network ID: 890
- Native currency: MARS
- Decimals: 18
- RPC: https://rpcs.marschain.net/
- Explorer: https://explorer.marschain.net/
- Website: https://marschain.net/
- GitHub: https://github.com/marschain-protocol/

### Chain ID reuse acknowledgement

We acknowledge that Chain ID 890 was previously used by a deprecated network. Therefore, this entry includes:

```json
"redFlags": ["reusedChainId"]
```

MarsChain has an independently running mainnet, genesis configuration, RPC infrastructure, and block explorer.

Please let us know if additional verification, genesis files, RPC tests, or proof of network operation are required.

PR #8600, opened 2026-08-16
Delete _data/chains/eip155-890.json

capitalexchange:
https://github.com/ethereum-lists/chains/pull/8578


ligi (changes requested):

PR #8632, opened 2026-08-23
Update L7 (717) to rpc.l7c.ai and add L7 rail RPCs.

sneakerkhan:
The old endpoint https://l7c.ai/rpc now 308s to https://rpc.l7c.ai. Chain name is L7; native symbol stays L7C.
Rail fronts at rpc.l7c.ai/l7c return that rail's 717


ligi (changes requested):
add your RPC to the end

PR #8689, opened 2026-09-08
Move Vordium mainnet from chainId 1110 to 101101

Vordium:
Vordium has relaunched on chainId 101101. This PR adds `eip155-101101.json` as the active chain and marks our old `eip155-1110.json` as deprecated.

Why the move: chainId 1110 collides with GRX Mainnet, which is listed at 1110 in DefiLlama's chainlist registry. Rather than contest an existing listing, we vacated 1110 and relaunched on 101101, which is unused both here and in DefiLlama's registry.

Per review, 1110 is **not** deleted — it keeps its file with `"status": "deprecated"` so the chainId stays permanently reserved and cannot be reclaimed. Its `rpc` and `explorers` are emptied because `rpc.vordium.com` and vordscan now serve 101101, and it is renamed to `"Vordium Legacy"` / `shortName: "vord-legacy"` so the active entry can keep the `vord` shortName.

RPC `https://rpc.vordium.com` now returns `0x18aed` (101101). Explorer is https://vordscan.io. The icon (`vordium`) is unchanged, so no icon file change is needed.

Previous entries: #8684 (mainnet) and #8685 (icon).


ligi (changes requested):

Vordium:
Thanks — understood, and that makes sense. Fixed.

`eip155-1110.json` is no longer deleted. It is restored with `"status": "deprecated"`, so the chainId stays permanently reserved and cannot be reclaimed by another chain.

Changes to the 1110 entry:
- `"status": "deprecated"` addedThanks, understood - fixed. 1110 is no longer deleted; it keeps its file with "status": "deprecated" so the chainId stays permanently reserved and cannot be reclaimed by another chain.
Changes to the 1110 entry: status set to deprecated, rpc and explorers emptied (rpc.vordium.com and vordscan now serve 101101, so leaving them would point at the wrong chain), and renamed to "Vordium Legacy" with shortName "vord-legacy" so the active entry can keep "vord".

eip155-101101.json is added as the active chain, so the PR is now add + deprecate rather than a rename. CI is green.
- `rpc` and `explorers` emptied, since `rpc.vordium.com` and vordscan now serve 101101 and would otherwise point at the wrong chain
- renamed to `"Vordium Legacy"` / `shortName: "vord-legacy"` so the new entry can keep the `vord` shortName

`eip155-101101.json` is added as the active chain. So the PR is now add + deprecate rather than a rename.

ligi (approved):

PR #8694, opened 2026-09-09
Tajirchain

MacromoduleTech:
Updated the ipfs logo width and height to 256 x 256 as the logo is not visible on the chainlist


ligi (approved):

MacromoduleTech:
@ligi I want to inquire if the logo on the chainlist will be updated . Currently I can see that the logo is not appearing on the chainlist

PR #8697, opened 2026-09-09
Fix/explorer dns resilience

Demon723:


ligi (changes requested):
ROFL - nope - remove this change to the explorer check

